PR #108, opened 2026-05-01
docs(skills): centralize preference gate prompts

stan4git:
## Summary

- New `plugin/skills/muggle-preferences/preference-gates.md` is the single source of truth for every preference-gated picker — Picker 1 questions/options, Picker 2 template, value mappings, silent-mode footer wording.
- Every consuming SKILL.md (`muggle-test`, `muggle-test-feature-local`, `muggle-test-import`, `muggle-test-regenerate-missing`, `muggle-status`, `muggle-pr-visual-walkthrough`) drops its inline gate boilerplate and references the central doc; per-step sections now state only which gate fires plus the step's side effects.
- `muggle-preferences` SKILL.md gains a "Change one" mode that handles `/muggle-preferences <key>` (the silent-mode footer command), re-running just that gate's Picker 1 and saving immediately.

## Picker UX fixes baked in

- No internal jargon in user-facing strings — e.g. "Browser window" instead of "Electron UI", no `showElectronBrowser = always` shown to the user.
- Sub-labels add information rather than restating the title.
- Picker 1 options sit on the same axis (e.g. *watch live* / *skip watching*, not *watch live* / *faster*).
- Picker 2 phrases the question in plain language ("Always run hidden from now on, without asking?") and saves the value that matches what the user picked — fixes the case where choosing "Run headless" would persist `showElectronBrowser = always`.

## `postPRVisualWalkthrough` is now PR-specific

Every prompt that targets a PR detects the PR first via `gh pr view` and substitutes the actual PR number/title into the question. If the branch has no open PR, the gate forks into a "create a PR or skip" picker (and refuses to silently auto-create even when the saved value is `always`).

## Silent-mode footer

Whenever a gate skips its picker because the user has saved `always` / `never`, the skill prints a short line under the auto-action:

```
✓ Continuing as user@example.com. Change with `/muggle-preferences autoLogin`.
```

The slash command points at the new "Change one" mode in `muggle-preferences`, giving the user a one-shot path to override.

## Test plan

- [ ] Open one of the consuming skills (e.g. `muggle-test-feature-local`) and confirm the `## Preferences` section is now ~5 lines with a pointer to `preference-gates.md`, not a duplicated 2-picker explanation.
- [ ] Trigger the `showElectronBrowser` gate via `/muggle-test-feature-local` with no saved preference — verify the picker reads "Browser window" / "Show the test browser as it runs?" with paired sub-labels, no `showElectronBrowser` variable name visible.
- [ ] Choose "Run hidden", then "Yes, always" — verify `showElectronBrowser` is saved as `never` (matching the choice), not `always`.
- [ ] Set `showElectronBrowser = always` globally, run the skill again — verify pickers are skipped and the silent-mode footer prints with `/muggle-preferences showElectronBrowser`.
- [ ] Run the printed command — verify the gate's Picker 1 re-runs and saving immediately persists the new value (no Picker 2 noise).
- [ ] Trigger `postPRVisualWalkthrough` on a branch with an open PR — verify the prompt names the specific PR (`#123 (title)`).
- [ ] Trigger it on a branch with no PR — verify Case B picker offers "Create a PR and post" / "Skip" instead of assuming a PR exists.


plugin/skills/muggle-preferences/SKILL.md
stan4git: Aren't these defined in the preference's definition?

Why read every perference in this definition?

stan4git: Stop repeating the values here, write the values into a centalized doc and scalefold here

stan4git: Addressed in c6e9f6b — moved all per-key descriptions/categories/allowed-values into the `Keys` table at the top of `preference-gates/README.md`. Configure now builds its questions from that index (groups by `Category`, uses `Description` column for option labels) instead of restating the same data in each question.

stan4git: Addressed in c6e9f6b — Set now validates `key` against the index in `preference-gates/README.md` and `value` against that key's `Allowed values` column. Same for Reset (iterates over the index). No more hardcoded 12-key list here.

plugin/skills/muggle-preferences/preference-gates/README.md
```diff
@@ -0,0 +1,65 @@
+# Preference Gates — Contract
+
+One file per gate in this directory. Skills load this contract + only the
+gates they actually fire.
+
+## Keys (single source of truth)
+
+12 total. The `muggle-preferences` skill renders its List, Configure, and
```
stan4git: You probably should just put a reference here. Why duplicate all these context for each keys here?

plugin/skills/muggle-preferences/preference-gates/README.md
```diff
@@ -0,0 +1,65 @@
...
+- `always` → take the pro-action, then print silent footer.
+- `never` → take the skip-action, then print silent footer.
+- `ask` (or absent) → run Picker 1 (per-key file) → Picker 2 (below).
+
+`defaultExecutionMode` uses `local`/`remote` instead of `always`/`never`.
+
+## Silent footer (whenever pickers are skipped)
+
+```
+✓ <silent action from per-key file>. Change with `/muggle-preferences <key>`.
```
stan4git: Please, write something human can understand. What the fuck do you mean here? No context?? how the fuck should a user knows what you did silently in the back???

PR #112, opened 2026-05-05
fix(muggle-preferences): natural-language picker pass + lean SKILL + 'ask each time'

stan4git:
## Summary

Three follow-ups to PR #111 — UX cleanup of the preference picker copy and a token-cost reduction on the skill itself.

### 1. Slim `SKILL.md` + trim MCP `preferences-set` response (`a8d8800`)

- `SKILL.md`: 124 → 29 lines. Now a dispatcher that points to one of five op files in `ops/` (change-one, list, configure, set, reset) based on user intent. The model loads only the file matching the requested op instead of the whole skill on every invocation.
- `muggle-local-preferences-set`: dropped the multi-line response that echoed every resolved preference + the file path. Returns the one-line `**<key>** set to **<value>** (<scope>).` only.
- Net: **Change one** dropped from ~4K to ~1.2K tokens per invocation.

### 2. 'Ask me each time' third option in Change-one picker (`2e0ce8d`)

The silent-mode footer points users at `/muggle-preferences <key>` to change a saved value, but the picker only offered the gate's two action options (`always`/`never`) — no way to revert to `ask`. Added a third `Ask me each time` option.

### 3. Natural-language pass on gate copy (`7d24b20`)

Read every gate file as a normal user would. Fixes:

- `autoLogin` silent: `Forcing a fresh login` → `Logging in fresh` (less aggressive).
- `autoSelectProject` description: dropped 'cached' jargon and the internal `<cwd>/.muggle-ai/last-project.json` path.
- `showElectronBrowser` description: dropped 'Electron' implementation detail.
- `openTestResultsAfterRun` silent: dropped redundant '— no auto-open' tail.
- `autoDetectChanges`: 'Yes/No I'll specify' → action-verb labels (`Scan changes`/`Skip scan`); fixed user-voice slip in sub-label.
- `suggestRelatedUseCases` / `suggestRelatedTestCases`: dropped 'Surface' corporate jargon; action-verb labels; fixed `I'll ask if I want them later` voice confusion.
- `checkForUpdates`: dropped 'Yes/No' framing; merged trailing 'Requires a network call.' into the question itself.
- `postPRVisualWalkthrough` Case A: `Yes, post to #N` → `Post to #N`.
- `postPRVisualWalkthrough` Case B fall-through: stopped leaking raw `postPRVisualWalkthrough = always` to user (violated the 'never put raw key in user-visible text' contract).

Action-verb labels also align with the new `Ask me each time` third option — `Yes/No` reads weird with `Ask` added.

## Test plan

- [ ] `/muggle-preferences` (no args) — List shows all 12 keys with current values + descriptions
- [ ] `/muggle-preferences <key>` — Change one fires Picker 1 from the gate file with three options including `Ask me each time`
- [ ] Picking `Ask me each time` saves `<key>=ask`
- [ ] Each gate's Picker 1 reads naturally — no jargon, no voice confusion, no raw key leaks
- [ ] `muggle-local-preferences-set` returns the one-line confirmation only
- [ ] `postPRVisualWalkthrough` Case B fall-through message no longer shows `postPRVisualWalkthrough = always` literal

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/plugin/skills/muggle-preferences/ops/change-one.md
+++ b/plugin/skills/muggle-preferences/ops/change-one.md
@@ -1,7 +1,7 @@
 # Change one — `/muggle-preferences <key>`
 
 1. Verify `preference-gates/<key>.md` exists. If not, list `preference-gates/*.md` and ask.
-2. Read it. Run **Picker 1** with **three options**: the gate's two options (mapped to `always`/`never`) plus `Ask me each time` (sub: `Prompt me at decision time.`) → `ask`.
+2. Read it. Run **Picker 1** with the options defined in the gate file, plus an additional `Ask me each time` (sub: `Prompt me at decision time.`) → `ask`.
 3. `muggle-local-preferences-set` with the mapped value, `scope: "global"`.
 4. Confirm: `Set <key> to <value>.`
 
--- a/plugin/skills/muggle-preferences/ops/change-one.md
+++ b/plugin/skills/muggle-preferences/ops/change-one.md
@@ -1,0 +1,8 @@
+# Change one — `/muggle-preferences <key>`
+
+1. Verify `preference-gates/<key>.md` exists. If not, list `preference-gates/*.md` and ask.
+2. Read it; run **Picker 1 only** per its spec.
+3. `muggle-local-preferences-set` with the mapped value, `scope: "global"`.
+4. Confirm: `Set <key> to <value>.`
+
+Skip Picker 2 — user explicitly asked to change.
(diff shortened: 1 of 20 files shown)


plugin/skills/muggle-preferences/ops/change-one.md
```diff
@@ -0,0 +1,10 @@
+# Change one — `/muggle-preferences <key>`
+
+1. Verify `preference-gates/<key>.md` exists. If not, list `preference-gates/*.md` and ask.
+2. Read it. Run **Picker 1** with **three options**: the gate's two options (mapped to `always`/`never`) plus `Ask me each time` (sub: `Prompt me at decision time.`) → `ask`.
```
stan4git: This should come from the option definition. Not always three?

stan4git: Good catch — fixed in dd9b0e1. Picker 1 options now come from the gate file (variable count, since e.g. `autoSelectProject`'s picker is the project list); change-one just adds one extra `Ask me each time` on top.

PR #115, opened 2026-05-05
feat(localDevHost): MCP cache tools + skill wiring

stan4git:
## Summary

Implements the cache infrastructure for the `localDevHost` gate (added in #112) and wires it into the two skills that ask for a local URL. Mirrors the existing last-project pattern.

### MCP plumbing

- `packages/mcps/src/shared/last-host.ts` — `read`/`write`/`clear`/`format` for `<cwd>/.muggle-ai/last-host.json` (versioned JSON, same shape as `last-project.json`)
- `last-host-schemas.ts` — Zod input schemas for the 3 tools
- 3 new MCP tools registered: `muggle-local-last-host-get`, `muggle-local-last-host-set`, `muggle-local-last-host-clear`
- `last-host.test.ts` — unit + schema tests

### Session context injection

- `ensure-electron-app.sh` injects a `Muggle Last Host: <url>` line, so skills read the cached URL with zero token cost — same trick the last-project cache uses.

### Skill wiring

- `muggle-test-feature-local` Step 4 (Local URL): gate `localDevHost` with three-way branch (`always` + cache → silent reuse / `never` → always picker / `ask` → Picker 1 + Picker 2). Always calls `last-host-set` after the pick so future runs see the latest URL.
- `muggle-test` Step 7A (pre-flight URL): same wiring.

### Effect

After the user opts to remember a URL once, subsequent runs in the same repo skip the URL prompt entirely and go straight to execution.

## Test plan

- [ ] `muggle-local-last-host-get` returns null on a fresh repo, returns the cached URL after `set`
- [ ] `muggle-local-last-host-set` creates `.muggle-ai/last-host.json` with the right shape
- [ ] `Muggle Last Host: <url>` line shows in session context after a write (verify in next session)
- [ ] `muggle-test-feature-local` with `localDevHost=ask`: Picker 1 shows last/suggested/type options
- [ ] `muggle-test-feature-local` with `localDevHost=always` + cache: URL reused silently with the silent-mode footer
- [ ] `muggle-test` Step 7A: same as above

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/plugin/scripts/ensure-electron-app.sh
+++ b/plugin/scripts/ensure-electron-app.sh
@@ -77,8 +77,9 @@ if [ -f "$prefs_global_file" ]; then
     try {
       const g = JSON.parse(fs.readFileSync('${prefs_global_file}', 'utf-8')).preferences || {};
       const defaults = {
-        autoLogin:'ask', autoSelectProject:'ask', showElectronBrowser:'ask',
-        openTestResultsAfterRun:'ask', defaultExecutionMode:'ask', autoPublishLocalResults:'ask',
+        autoLogin:'ask', autoSelectProject:'ask', autoSelectLocalHost:'ask',
+        showElectronBrowser:'ask', openTestResultsAfterRun:'ask',
+        defaultExecutionMode:'ask', autoPublishLocalResults:'ask',
         suggestRelatedUseCases:'ask', suggestRelatedTestCases:'ask', autoDetectChanges:'ask',
         postPRVisualWalkthrough:'ask', checkForUpdates:'ask', verboseOutput:'ask'
       };
@@ -126,7 +127,7 @@ fi
 # --- Last-host cache injection ---
 # Per-repo cache of the local dev server URL the user picked on the previous
 # run. Lives at <cwd>/.muggle-ai/last-host.json. Skills silently reuse it
-# when the user has set the localDevHost preference to "always".
+# when the user has set the autoSelectLocalHost preference to "always".
 last_host_line=""
 last_host_note=""
 last_host_line=$(node -e "
--- a/plugin/scripts/ensure-electron-app.sh
+++ b/plugin/scripts/ensure-electron-app.sh
@@ -90,18 +90,18 @@ if [ -f "$prefs_global_file" ]; then
       const hasProject = Object.keys(p).length > 0;
       const note = hasProject ? ', project overrides active' : '';
       const line = Object.entries(merged).map(([k,v]) => k+'='+v).join(' ');
-      console.log('Muggle Preferences (~/.muggle-ai/preferences.json' + note + '):\\\\n' + line);
+      console.log('Muggle Test Preferences (~/.muggle-ai/preferences.json' + note + '):\\\\n' + line);
     } catch { console.log(''); }
   " 2>/dev/null || true)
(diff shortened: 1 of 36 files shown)


plugin/scripts/ensure-electron-app.sh
```diff
@@ -123,7 +123,29 @@ if [ -n "$last_project_line" ]; then
...
+# --- Last-host cache injection ---
+# Per-repo "last used local dev server URL" cache. Lives at <cwd>/.muggle-ai/last-host.json
+# and is honored by skills when localDevHost = always.
```
stan4git: Why this is "always"?? Shouldn't this be the host user provided?

This is so confusing

plugin/skills/muggle-test-feature-local/SKILL.md
```diff
@@ -94,13 +95,16 @@ Before detecting the local URL, verify that the services the user needs are actu
 
 This step is especially important when the user's app depends on sibling services (a backend API, an auth service, etc.) that may not be running yet. The prepare skill handles discovery, startup, and cleanup so this skill doesn't have to.
 
-### 4. Local URL
+### 4. Local URL (gated by `localDevHost`)
```
stan4git: Shouldn't this be referencing a centralized md for this gate?

plugin/skills/muggle-test-feature-local/SKILL.md
```diff
@@ -29,9 +29,10 @@ Gates run per `preference-gates/README.md`.
 | Preference | Step | Decision it gates |
 |------------|------|-------------------|
 | `autoLogin` | 1 | Reuse saved credentials when auth is required |
-| `autoSelectProject` | 2 | Reuse last-used Muggle project for this repo |
+| `autoSelectProject` | 2 | Reuse last-used Muggle Test project for this repo |
+| `localDevHost` | 4 | Reuse last-used local dev server URL for this repo |
```
stan4git: This is not a boolean right??

If it is a boolean, name it like a boolean.
If it is a value, name it like a value

plugin/scripts/ensure-electron-app.sh
stan4git: Fixed in 8c06123→latest. Reworded the comment so 'always' clearly refers to the saved preference value, not 'every time'. Cache itself stores the URL the user picked.

plugin/skills/muggle-test-feature-local/SKILL.md
stan4git: Fixed in 8c06123→latest. The two Local-URL sections (here and in muggle-test Step 7A) now only describe skill-side concerns (read cache / auto-detect / save) and reference `preference-gates/localDevHost.md` for picker options, branch logic, and the save-on-every-pick rule. Same shape as `autoSelectProject`'s wiring earlier in this file.

plugin/skills/muggle-test-feature-local/SKILL.md
stan4git: Fixed in this push. Renamed `localDevHost` -> `autoSelectLocalHost` everywhere — direct parallel to `autoSelectProject` and matches the existing `auto*` convention for boolean gates. The name now clearly signals it's a behavior toggle (silently reuse the saved URL or not), not a value.

PR #123, opened 2026-05-08
Skill gate eval — runtime + first signal

stan4git:
## Summary

Wires the Layer 2 harness against the real Claude Agent SDK and runs the first behavioral evals for `showElectronBrowser` × `muggle-test-feature-local`.

## What's in this PR

- `@anthropic-ai/claude-agent-sdk` added as a devDependency.
- `mock-mcp.ts` rewritten to build the MCP server via `createSdkMcpServer` + `tool()` (the SDK's in-process MCP path).
- `harness.ts` calls the SDK's `query()` with mock MCP wired in, intercepts `AskQuestion` via `canUseTool`, and records every tool call.
- Mock server is named `eval_mock` so it doesn't collide with the production `plugin_muggle_muggle` namespace that the SDK inherits from the user's Claude Code settings. Production-prefix calls get denied with a redirect message; an ENVIRONMENT NOTE in the system prompt tells the agent the same thing in prose.
- `EXECUTE_SHAPE` declares the real arg shape on `muggle-local-execute-*` because zod strips undeclared keys before canUseTool sees them — without this, the agent's `showUi`/`freshSession` are invisible to the verdict.
- `run.ts` gains `--scenario <substring>` filter and `--verbose` for cheap iteration.

## First results

Sonnet 4.6, 1 rep per scenario:
- `always-omits-showUi`: **PASS**
- `never-passes-showUi-false`: **PASS**
- `absent-fires-picker1`: **FAIL**

The third failure is the actual gate misfire — the agent's thinking literally says "`showElectronBrowser` preference - not set in the session context" and then calls `muggle-local-execute-test-generation` directly without firing Picker 1. The skill prose treats absent the same as "use default" instead of "ask," contradicting the gate contract (`absent → ask`). Fix is upstream of this PR — likely a SKILL.md tweak or a clearer preference-gates README.

Results landed in `muggle-ai-brain/eval/skill-gate-eval/showElectronBrowser/results.json`.

## Test plan

- [x] `pnpm tsc --noEmit -p internal/skill-gate-eval/tsconfig.json` — clean
- [x] `pnpm vitest run src/test/skills/preference-gates-lint.test.ts` — 49/49 still green
- [x] One rep of each `showElectronBrowser` scenario — runs end-to-end, surfaces the real misfire
- [ ] Reviewer: skim the system-prompt synthesis at `harness.ts` and the canUseTool dispatch

### DIFF
--- a/plugin/skills/muggle-test-feature-local/SKILL.md
+++ b/plugin/skills/muggle-test-feature-local/SKILL.md
@@ -158,8 +158,9 @@ The MCP client often uses a **default wait of 300000 ms (5 minutes)** for `muggl
 Call `muggle-local-execute-test-generation` or `muggle-local-execute-replay` directly. **Do not** ask the user to re-approve the Electron launch — the user choosing this skill in the first place is the approval.
 
 Gate `showElectronBrowser` (per `preference-gates/README.md`). Reuse choice within a session.
-- Pro-action: omit `showUi`.
-- Skip-action: pass `showUi: false`.
+- `always` → omit `showUi`.
+- `never` → pass `showUi: false`.
+- `ask` → **invoke the `AskUserQuestion` tool — not a prose question, not markdown, not a final assistant message that asks**. The Picker 1 contract lives in `preference-gates/showElectronBrowser.md` (header `Browser window`, question `"Show the test browser as it runs?"`, options `Show it` / `Run hidden`). Call `AskUserQuestion` with that header/question/options BEFORE calling execute. Map the answer: `Show it` → omit `showUi`; `Run hidden` → pass `showUi: false`. If you find yourself writing the question in regular text instead, stop and use the tool.
 
 ### 8. After successful generation only (open `viewUrl` gated by `openTestResultsAfterRun`)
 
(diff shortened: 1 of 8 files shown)


plugin/skills/muggle-test-feature-local/SKILL.md
```diff
@@ -158,8 +158,9 @@ The MCP client often uses a **default wait of 300000 ms (5 minutes)** for `muggl
 Call `muggle-local-execute-test-generation` or `muggle-local-execute-replay` directly. **Do not** ask the user to re-approve the Electron launch — the user choosing this skill in the first place is the approval.
 
 Gate `showElectronBrowser` (per `preference-gates/README.md`). Reuse choice within a session.
-- Pro-action: omit `showUi`.
-- Skip-action: pass `showUi: false`.
+- `always` → omit `showUi`.
+- `never` → pass `showUi: false`.
+- `ask` → **invoke the `AskUserQuestion` tool — not a prose question, not markdown, not a final assistant message that asks**. The Picker 1 contract lives in `preference-gates/showElectronBrowser.md` (header `Browser window`, question `"Show the test browser as it runs?"`, options `Show it` / `Run hidden`). Call `AskUserQuestion` with that header/question/options BEFORE calling execute. Map the answer: `Show it` → omit `showUi`; `Run hidden` → pass `showUi: false`. If you find yourself writing the question in regular text instead, stop and use the tool.
```
stan4git: why do you need to ramble about this?? keep it clean

PR #124, opened 2026-05-08
feat: /muggle-feedback skill + MCP user-feedback tools

stan4git:
## Summary
- New `/muggle-feedback` skill: routes submit / list / delete operations on user feedback against generated action scripts. Auto-detects the run anchor from a recent local run, a remote run, or a Muggle dashboard URL pasted in the prompt; renders the script for review and collects step-level + whole-script feedback in one batch.
- Three new MCP tools backing the skill: `muggle-remote-user-feedback-{create,list,delete}` mapped to `/v1/protected/muggle-test/user-feedback`.
- Chain hooks in `/muggle-test` and `/muggle-test-feature-local` that suggest feedback when a run fails or looks wrong, passing the run id as anchor.

## Design
Spec: `muggle-ai-brain/product/2026-05-08-muggle-feedback-skill-design.md` (Status: Pending).

## Notes
- `studioAuthInfo` is intentionally absent from the MCP create schema — the controller reads it from request headers and the feedback-analysis workflow falls through gracefully when it's missing (feedback is still saved).
- List uses `limit`/`offset` to match the existing controller signature (not `page`/`pageSize`).
- **Open issue worth tracking server-side:** the regeneration prompt does not currently consume the feedback text — only similarity matching does. So submitting feedback may regenerate the same script. Worth fixing before the skill is widely advertised.

## Test plan
- [x] `pnpm exec tsc --noEmit` clean for `packages/mcps`
- [ ] CI green
- [ ] Smoke-test from a Claude Code session: `/muggle-feedback` after a local run → submit → confirm feedback row + analysis runtime id surfaced
- [ ] Smoke-test list/delete paths

### DIFF
--- a/packages/mcps/src/mcp/tools/e2e/tool-registry.ts
+++ b/packages/mcps/src/mcp/tools/e2e/tool-registry.ts
@@ -1424,7 +1424,7 @@ const userFeedbackTools: IQaToolDefinition[] = [
   {
     name: "muggle-remote-user-feedback-list",
     description:
-      "List active user feedback entries for a project. Supports limit/offset pagination. Response: { feedback: IUserFeedback[], total: number, hasMore: boolean }.",
+      "List active user feedback entries for a project. Optionally narrow by exactly one of actionScriptId / testScriptId / testCaseId / useCaseId — the typical query is 'show me feedback on this test case (or test script, or use case).' Supports limit/offset pagination. Response: { feedback: IUserFeedback[], total: number, hasMore: boolean }.",
     inputSchema: schemas.UserFeedbackListInputSchema,
     mapToUpstream: (input) => {
       const data = input as z.infer<typeof schemas.UserFeedbackListInputSchema>;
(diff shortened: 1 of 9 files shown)


packages/mcps/src/mcp/tools/e2e/tool-registry.ts
```diff
@@ -1398,6 +1398,61 @@ const apiKeyTools: IQaToolDefinition[] = [
...
+          target: data.target,
+          feedbackText: data.feedbackText,
+        },
+      };
+    },
+  },
+  {
+    name: "muggle-remote-user-feedback-list",
+    description:
+      "List active user feedback entries for a project. Supports limit/offset pagination. Response: { feedback: IUserFeedback[], total: number, hasMore: boolean }.",
```
stan4git: I think mostly I will request feedbacks of a user case/test case/test script

stan4git: Good call — fixed in e4b8223 (works) + multiplex-ai/muggle-ai-prompt-service#475 (server).

The list tool now accepts optional, mutually-exclusive `actionScriptId` / `testScriptId` / `testCaseId` / `useCaseId` filters; `projectId` stays required as the auth boundary. Server resolves the narrowing filter to a list of action-script ids inside the project, fans out to the existing `listUserFeedbackByActionScriptIdAsync` DAO, merges, sorts, paginates. Skill `list.md` was rewritten to pick a scope first (test case is the default suggestion) instead of post-filtering client-side.

This works PR depends on the server PR being deployed first.

PR #127, opened 2026-05-10
Three-branch enumeration on every gate stanza

stan4git:
## Summary

Follow-up to #125 — the trim commits that addressed your "keep it clean" review feedback (`5ae6c4c`, `2de50eb`) were pushed AFTER the merge and never made it to master. This PR is just those two commits squashed into a single clean change.

Replaces the previous two-branch "Pro-action / Skip-action" stanzas with explicit three-branch enumeration:

- `always` → pro-action
- `never` → skip-action
- `ask` → run Picker 1 from `preference-gates/<key>.md` via `AskUserQuestion`; map the answer back to one of the actions above.

The eval in #123 showed the LLM doesn't reliably consult the README's "absent → ask" rule from a two-branch stanza — when the resolved value was `ask` the agent skipped the gate. Spelling out the third branch with the tool name named at the call site pins the behavior.

## Touched gates

`autoLogin` (×4 sites), `autoDetectChanges`, `showElectronBrowser` (×2), `openTestResultsAfterRun`, `autoPublishLocalResults`, `checkForUpdates`, `suggestRelatedUseCases`, `suggestRelatedTestCases`.

## Test plan

- [x] `pnpm vitest run src/test/skills/preference-gates-lint.test.ts` — 98/98 (Layer 1 contract lint)
- [x] No `Pro-action` / `Skip-action` strings remain anywhere in `plugin/skills/`
- [x] Behavioral eval on #123 branch showed 3/3 PASS at 3 reps each on Sonnet 4.6 with this same stanza shape

### DIFF
--- a/plugin/skills/muggle-test-regenerate-missing/SKILL.md
+++ b/plugin/skills/muggle-test-regenerate-missing/SKILL.md
@@ -52,8 +52,9 @@ Treat this filter as a default, not a law. If the user explicitly says "include 
 
 1. Call `muggle-remote-auth-status`.
 2. If **authenticated and not expired** → gate `autoLogin` (per `preference-gates/README.md`):
-   - Pro-action: proceed with saved session.
-   - Skip-action: `muggle-remote-auth-login` with `forceNewSession: true`, then `muggle-remote-auth-poll`.
+   - `always` → proceed with saved session.
+   - `never` → `muggle-remote-auth-login` with `forceNewSession: true`, then `muggle-remote-auth-poll`.
+   - `ask` → run Picker 1 from `preference-gates/autoLogin.md` via `AskUserQuestion`; map the answer back to one of the actions above.
 3. If **not authenticated or expired** → call `muggle-remote-auth-login`, then poll with `muggle-remote-auth-poll`.
 4. Do not skip auth and do not assume a stale token still works.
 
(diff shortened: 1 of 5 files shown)


plugin/skills/muggle-test-regenerate-missing/SKILL.md
stan4git, line 1: seems a redundant skill? who and where is this called?

stan4git:
On the `muggle-test-regenerate-missing` redundancy question — what I found:

- It's user-invocable via `/muggle:muggle-test-regenerate-missing` directly.
- Listed in `plugin/README.md` and root `README.md` as a user-facing command.
- Underlying MCP tool has tests in `packages/mcps/src/test/bulk-test-script-generation.test.ts`.
- **But** the muggle router (`plugin/skills/muggle/SKILL.md`) doesn't reference it in either its menu or its routing rules — so `/muggle` with no subcommand won't surface it. Only the explicit slash command path works.

That mismatch (listed in docs, missing from the router) suggests the skill is half-promoted. Whether to fully promote (add to router) or deprecate (remove the skill + docs entries) is a product call I shouldn't make in this PR.

Dropped the `muggle-test-regenerate-missing/SKILL.md` change from this PR so the rest of the gate-stanza work isn't blocked on that decision. PR is now scoped to the four actively-used skills only. Force-pushed.

PR #131, opened 2026-05-10
chore: remove muggle-test-regenerate-missing skill

stan4git:
## Summary

Per the #127 review feedback ("seems a redundant skill? who and where is this called?"), removing `muggle-test-regenerate-missing` rather than promoting it.

The skill was half-promoted — listed in user-facing READMEs as a slash command but missing from the `/muggle` router's menu and routing rules, so it was only reachable via the explicit slash invocation. Going with cleanup.

## Removed

- `plugin/skills/muggle-test-regenerate-missing/` (SKILL.md + 2 eval files, 305 lines)
- Row in `plugin/README.md`'s skills table
- Bullet in root `README.md`'s install list
- Tree mention in root `README.md`'s directory tree
- Reference in `plugin/skills/_shared/failure-mode-handling.md`'s "source of truth" comment

## Untouched (intentional)

- The underlying MCP tool `muggle-remote-workflow-start-test-script-generation-bulk` and its tests in `packages/mcps/src/test/bulk-test-script-generation.test.ts`. The bulk-generation API is still callable directly; only the skill that surfaced it is gone. The `"muggle-test-regenerate-missing: bulk (...)"` display name in the test fixtures stays as historical naming — doesn't affect behavior.

## Test plan

- [x] `pnpm vitest run src/test/skills` — 96/96 (was 98/98; two assertions disappeared with the gates this skill referenced)
- [x] `pnpm vitest run packages/mcps/src/test/bulk-test-script-generation.test.ts` — 13/13
- [x] `grep -rln "muggle-test-regenerate-missing" --include="*.md"` (excluding worktrees) returns no hits


stan4git:
On reflection: this skill is the only user-facing path for 'scan a project, filter DRAFT/GENERATION_PENDING test cases, bulk-regen the missing ones.' The bulk-generation MCP tool alone is just the dispatch — no scan/filter/confirm. Removing it loses real functionality. Keeping the skill; the missing-router-mention is a follow-up concern, not justification for deletion.

PR #135, opened 2026-05-11
feat(skills): short m* aliases for muggle-* skills and slash commands

stan4git:
## Summary

- Adds 14 short-name aliases for the muggle-* skills (router + 13 leaf skills) so users can invoke them with far fewer keystrokes — e.g. `/mtest` instead of `/muggle-test`, `/mdo` instead of `/muggle-do-task`.
- Each alias ships in two forms for cross-IDE coverage:
  - **`plugin/commands/<alias>.md`** — Claude Code slash commands (thin wrappers that invoke the underlying skill via the `Skill` tool).
  - **`plugin/skills/<alias>/SKILL.md`** — alias skills with deliberately minimal descriptions so they only trigger on explicit name match (no auto-fire), picked up by Cursor via the existing skill sync.
- Updates the Cursor sync allowlist (`scripts/postinstall.mjs`) and the cleanup filter (`src/cli/cleanup.ts`) so the new `m*` names sync alongside the existing `muggle-*` prefix match. The two allowlists must stay in sync — flagged in code comments.

## Alias map

| Alias | Underlying skill |
|---|---|
| `m` | `muggle` (router) |
| `mtest` | `muggle-test` |
| `mdo` | `muggle-do-task` |
| `mpr` | `muggle-pr-visual-walkthrough` |
| `mprefs` | `muggle-preferences` |
| `mstatus` | `muggle-status` |
| `mrepair` | `muggle-repair` |
| `mupgrade` | `muggle-upgrade` |
| `mfeedback` | `muggle-feedback` |
| `mimport` | `muggle-test-import` |
| `mtestlocal` | `muggle-test-feature-local` |
| `mtestprep` | `muggle-test-prepare` |
| `mregen` | `muggle-test-regenerate-missing` |
| `mrelease` | `muggle-works-npm-release` |

## Test plan

- [x] `npx vitest run` — 98 tests pass (incl. `preference-gates-lint.test.ts`, which walks every skill dir and tolerates aliases with no Preferences table)
- [x] `npx tsc --noEmit` — clean
- [ ] Manual: `/mstatus` in Claude Code invokes `muggle-status`
- [ ] Manual: `/mtest` invokes `muggle-test` and forwards `\$ARGUMENTS`
- [ ] Manual: after `npm install -g @muggleai/works`, the m* alias dirs land in `~/.cursor/skills/`

### DIFF
--- a/src/cli/cleanup.ts
+++ b/src/cli/cleanup.ts
@@ -20,6 +20,24 @@ const CURSOR_SKILLS_SUBDIR = "skills";
 /** Prefix for muggle skills. */
 const MUGGLE_SKILL_PREFIX = "muggle";
 
+/** Short-name alias skills that ship alongside the muggle-* skills (kept in sync with scripts/postinstall.mjs). */
+const MUGGLE_ALIAS_SKILLS = new Set<string>([
+  "m",
+  "mtest",
+  "mdo",
+  "mpr",
+  "mprefs",
+  "mstatus",
+  "mrepair",
+  "mupgrade",
+  "mfeedback",
+  "mimport",
+  "mtestlocal",
+  "mtestprep",
+  "mregen",
+  "mrelease",
+]);
+
 /** Install manifest filename. */
 const INSTALL_MANIFEST_FILE = "install-manifest.json";
 
@@ -145,7 +163,7 @@ export function listObsoleteSkills (): IObsoleteSkill[] {
         continue;
       }
 
-      if (!entry.name.startsWith(MUGGLE_SKILL_PREFIX)) {
+      if (!entry.name.startsWith(MUGGLE_SKILL_PREFIX) && !MUGGLE_ALIAS_SKILLS.has(entry.name)) {
         continue;
       }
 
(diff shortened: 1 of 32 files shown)


src/cli/cleanup.ts
```diff
@@ -20,6 +20,24 @@ const CURSOR_SKILLS_SUBDIR = "skills";
 /** Prefix for muggle skills. */
 const MUGGLE_SKILL_PREFIX = "muggle";
 
+/** Short-name alias skills that ship alongside the muggle-* skills (kept in sync with scripts/postinstall.mjs). */
+const MUGGLE_ALIAS_SKILLS = new Set<string>([
```
stan4git: Why can't this be a shared static file?

stan4git: Good call — extracted to `plugin/skills/_aliases.json` in beafff7. Both `scripts/postinstall.mjs` and `src/cli/cleanup.ts` now load from the same manifest. Cleanup uses an upward walk from `import.meta.url` so the lookup works in both source (vitest/tsx) and the bundled `dist/` output.

PR #136, opened 2026-05-11
feat(skills): add branch-hygiene guidance to dev/test skills

stan4git:
## Summary

- New `_shared/branch-hygiene.md` as the single source of truth for three recommended habits: use a worktree for non-trivial dev work, rebase onto the default branch before running dev servers or E2E tests, and clean up after a PR is merged (worktree, branch, local Muggle Test artifacts, `commit-commands:clean_gone`).
- Wired the doc into `muggle-test-feature-local`, `muggle-test-prepare`, and `muggle-do` (plus its `do/pre-flight.md` and `do/open-prs.md` stages) so each skill picks up the relevant section at the right point in its workflow.
- All three recommendations surface via `AskUserQuestion` with explicit opt-out — nothing is auto-applied, no destructive git command runs without confirmation.

## Test plan

- [ ] Trigger `muggle-test-feature-local` on a branch that is behind `origin/master` and confirm the rebase prompt fires before the Execute step.
- [ ] Trigger `muggle-test-prepare` on a stale branch and confirm Step 0 surfaces the rebase recommendation before any servers start.
- [ ] Run `muggle-do` end-to-end and verify pre-flight collects worktree (Q11) and rebase (Q12) choices in the same consolidated turn.
- [ ] Confirm `do/open-prs` finishes by pointing the user at the cleanup commands in `_shared/branch-hygiene.md` section 3 without running them.


plugin/skills/_shared/branch-hygiene.md
stan4git, line 1: This should be broken down into files to reference only the parts needed.

stan4git: Addressed in 6c73538 — split into three flat files in `_shared/`:

- `use-worktrees.md`
- `rebase-before-e2e.md`
- `post-merge-cleanup.md`

Each caller now references only what it needs (e.g. `muggle-test-prepare` and `do/open-prs.md` link to just one file). The `section 1/2/3` references are gone, so each link is self-describing.

PR #140, opened 2026-05-11
chore(skills): centralize _shared content, fix branch-hygiene refs

stan4git:
## Summary

Follow-up from the five-PR session (#136, #139, #137, #138, #132). Stan flagged two issues with what landed:

1. `_shared/` had overlapping files (`use-worktrees.md` and `worktree-isolation.md`) that weren't cross-linked.
2. Callers were re-stating instructions that already lived in canonical `_shared/` docs instead of just referencing them.

This PR centralizes the duplicated content and fixes two dead references to the deleted `branch-hygiene.md`.

### Worktree docs merged
- **Deleted** `_shared/use-worktrees.md` (was 7 lines, the `autoUseWorktree` gate-procedure stub).
- **Folded its content** into `_shared/worktree-isolation.md` as a new top section. The file now leads with the single-PR gate (when to create a worktree), then the multi-branch case, then the setup mechanics that already lived there.
- All callers updated (`do/pre-flight.md`, `muggle-do/SKILL.md`, `muggle-test-feature-local/SKILL.md`).

### Dead `branch-hygiene.md` refs fixed
- `_shared/worktree-isolation.md:14` — old line about \"single one-off change: see branch-hygiene.md section 1\" was removed by the merge (the new top section is the single-PR doc itself).
- `muggle-test/SKILL.md:545` — removed the bullet pointing at \`_shared/branch-hygiene.md sections 1A / 1B / 1C\` (that content already lives in `worktree-isolation.md` post-merge).

### Ready-signal patterns deduped (cut 4 re-statements)
Replaced re-stated \`Compiled successfully\` / \`ready in\` / \`ready - started server\` blocks with one-line references to \`_shared/dev-server-readiness.md\`:

| File | Section |
|------|---------|
| `do/e2e-acceptance.md` | Step 0.5 Probe 1.5 |
| `muggle-test-prepare/SKILL.md` | Startup verification (two-stage) |
| `muggle-test-feature-local/SKILL.md` | Compile-gate |
| `muggle-test/SKILL.md` | PR-loop step 6 |

### Port-kill commands deduped (cut 2 re-statements)
Replaced inline \`Get-NetTCPConnection\` / \`lsof -ti\` with references to the canonical helpers in \`_shared/worktree-isolation.md\`:

| File | Section |
|------|---------|
| `muggle-test-prepare/SKILL.md` | Option 2 — force-kill |
| `muggle-test/SKILL.md` | PR-loop step 4-5 |

### Cross-linked the two PR-creation gates
[...]

PR #132, opened 2026-05-10
feat(skills): muggle-do worktree-by-default + autoE2ETest pref gate

stan4git:
## Summary

Two coupled changes to `/muggle-do` (and a small one to `/muggle-test-feature-local`):

- **Worktree-by-default.** Pre-flight now creates (or reuses) a sibling git worktree at `<repo>-worktrees/<slug>` and records `worktreePath` per repo in `state.md`. Stages 2-7 operate exclusively inside that worktree, so the user's primary checkout is untouched throughout the autonomous cycle.
- **`autoE2ETest` preference gate.** New preference (default `always` -- the only pref where unset != ask) controls whether Stage 6 runs. When the value is `ask`, the question is folded into the consolidated pre-flight questionnaire (item 11) so we never break the "no mid-cycle questions" rule. When `never`, Stage 6 skips and the PR title gets `[E2E SKIPPED]`. The legacy `Validation: unit-only / skip` field becomes informational; only this pref controls run/skip.
- **Dev-server-on-the-right-branch verification.** Both Stage 6 and `/muggle-test-feature-local` now resolve the dev server's PID -> cwd -> branch and halt when it does not match the worktree/branch under test. Validating stale code is worse than not validating at all.

## Files changed

| File | Change |
| :--- | :--- |
| `plugin/skills/muggle-do/SKILL.md` | New "Worktree-based execution" + "E2E acceptance gate" sections; front-loading list updated |
| `plugin/skills/do/pre-flight.md` | Silent worktree-existence probe + dev-server cwd capture + autoE2ETest resolution; questionnaire item 11; new "Worktree creation" silent step; `state.md` schema records `original=` / `worktree=` per repo + `autoE2ETest`; two new non-negotiables |
| `plugin/skills/do/e2e-acceptance.md` | Step 0 rewritten around the pref gate; Step 0.5 adds dev-server cwd verification with halt-on-mismatch; two new non-negotiables |
| `plugin/skills/do/open-prs.md` | `git push` runs from `worktreePath`; skip-tag table now includes `[E2E SKIPPED]` |
| `plugin/skills/muggle-test-feature-local/SKILL.md` | New step 4a (PID -> cwd -> branch verification, halts on mismatch) + matching non-negotiable |
| `plugin/skills/muggle-preferences/SKILL.md` | `autoE2ETest` added to the table (default `always`), to the Set validation list, and to Reset (resets to `always`, not `ask`) |

## Caveat (heads-up for follow-up)

[...]

PR #141, opened 2026-05-12
chore(skills): extract walkthrough-or-create-pr.md + drop autoE2ETest from picker

stan4git:
## Summary

Follow-up to #140, addressing the two items I deferred from that PR. **Stacked on `chore/centralize-shared` — review/merge #140 first.**

### 1. Extract \`_shared/walkthrough-or-create-pr.md\`

\`muggle-test/SKILL.md\` (step 9b + 9c) and \`muggle-test-feature-local/SKILL.md\` (step 10b + 10c) had near-identical 10-line blocks describing the same procedure:

1. Detect the PR via \`gh pr view\`
2. Fire the \`postPRVisualWalkthrough\` gate (Case A or Case B)
3. Invoke \`muggle-pr-visual-walkthrough\` in Mode A

The bodies differed only in step numbering. Both files now reference the shared procedure in a single line.

### 2. Drop \`autoE2ETest\` from \`muggle-preferences/ops/configure.md\` picker

The configure picker is built around the multi-select pattern *\"Which of these should auto-proceed (set to always)?\"*. \`autoE2ETest\` doesn't fit:

- Its default is already \`always\`
- Its only override is \`ask\` (no \`never\`)
- Multi-select toggle would have to mean \"turn off the default\" rather than \"set to always,\" inverting the picker's mental model

Removed the standalone \`multiSelect: false\` category I'd shoehorned in during #132 and added a one-line inline note in \`configure.md\` explaining the exclusion so a future maintainer doesn't re-add it. Users who want \`ask\` still set it directly via \`/muggle-preferences autoE2ETest\`.

## Diff stat

\`\`\`
4 files changed, 38 insertions(+), 33 deletions(-)
\`\`\`

Net 5 lines removed even with the new shared file added — the dedup saved more than the extraction cost.

## Test plan

- [ ] \`pnpm test src/test/skills/preference-gates-lint.test.ts\` — symmetry should still hold; no gate file changes
- [ ] Spot-check: open \`_shared/walkthrough-or-create-pr.md\` and read end-to-end as the canonical procedure
- [ ] Grep \`grep -rn 'Detect the PR, then apply' plugin/\` — should return nothing (both old blocks removed)
- [ ] Open \`muggle-preferences/ops/configure.md\` and confirm no \`autoE2ETest\` line; confirm the inline exclusion note explains why


plugin/skills/_shared/walkthrough-or-create-pr.md
```diff
@@ -0,0 +1,32 @@
+# Post Walkthrough (or Create the PR First)
+
+Shared procedure for posting a Muggle Test visual walkthrough to a PR after a test run. Used by `muggle-test` (step 9b/9c) and `muggle-test-feature-local` (step 10b/10c).
```
stan4git: No need for this dependency

plugin/skills/_shared/walkthrough-or-create-pr.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Procedure
+
+### 1. Detect the PR
+
+```bash
+gh pr view --json number,title,url 2>/dev/null
+```
+
+Branch on the result into Case A or Case B.
```
stan4git: this is confusing, make you you point to case A or B in procedure 2

stan4git: Why mix the "post report" and "create pr"???

I don't like how you organize these. make sure you have a clean sepration of concern

plugin/skills/muggle-preferences/ops/configure.md
```diff
@@ -30,10 +30,11 @@ For each option: label = key name, description = first paragraph of `preference-
 - `multiSelect: true`, `header: "Test run"` — `showElectronBrowser`, `openTestResultsAfterRun`, `autoPublishLocalResults`
 - `multiSelect: true`, `header: "Suggestions & PR"` — `suggestRelatedUseCases`, `suggestRelatedTestCases`, `postPRVisualWalkthrough`, `autoCreatePR`
 - `multiSelect: true`, `header: "Branch hygiene"` — `autoUseWorktree`, `autoRebase`, `autoCleanup`
-- `multiSelect: false`, `header: "E2E acceptance"` — `autoE2ETest`. Options: `Always run Stage 6 at the end` (`always` — default), `Ask each cycle` (`ask`). No `never` option.
 - `multiSelect: false`, `header: "Default mode"` — `defaultExecutionMode`. Options: `Local — run on my computer` (`local`), `Remote — run in the Muggle Test cloud` (`remote`), `Ask each time` (don't change).
 - `multiSelect: false`, `header: "Scope"` — final scope question. Options: `Global (all repos)` (~/.muggle-ai/), `This project only` (.muggle-ai/ in repo).
 
+`autoE2ETest` is intentionally excluded — its default is already `always` and its only override is `ask`, so the "Which should auto-proceed (set to always)?" multi-select pattern doesn't fit. Users who want `ask` instead set it directly via `/muggle-preferences autoE2ETest`.
```
stan4git: why put this irrelavant info here. this is not generic

plugin/skills/_shared/walkthrough-or-create-pr.md
stan4git: Deleted the file entirely (per your other comments). The dependency goes away.

plugin/skills/_shared/walkthrough-or-create-pr.md
stan4git: File deleted (per your Case A/B + separation-of-concerns comments). Each caller now has its own inline 4-step orchestration with explicit gate firing — no Case A/B narration since the gates are pure post/skip + create/skip.

stan4git: Agreed — mixing was wrong. Deleted `_shared/walkthrough-or-create-pr.md`. Now: `postPRVisualWalkthrough` gate = post/skip only; `autoCreatePR` gate = create-PR only. Each caller (muggle-test step 9b, muggle-test-feature-local step 10b) chains the two gates explicitly in 4 lines.

plugin/skills/muggle-preferences/ops/configure.md
stan4git: Removed the exclusion note.

plugin/skills/muggle-test/SKILL.md
```diff
@@ -73,9 +73,7 @@ Signs the user wants this: mentions "preview", "staging", "deployed", "preview U
 >
 > Trigger phrases: "test all open PRs", "regression-test every open PR", "muggle-test on all the active PRs", pre-merge audit, scheduled cron.
 
-Mode C is **not** about testing the current branch. It's an orchestration loop that builds a per-PR worktree, starts a dev server in it, dispatches the `acceptance-tester` subagent, and tears down — once per PR, sequentially, because all PRs share a single dev-server port.
-
-Skip Mode C's full procedure (below) when the user just wants to test their current changes (use Mode A or B).
+Mode C is **not** about the current branch — it's an across-PRs orchestration loop. For "test my changes", use Mode A or B.
```
stan4git: Why?? "my changes" can be multiple PRs?

plugin/skills/muggle-test-feature-local/SKILL.md
```diff
@@ -223,20 +223,12 @@ After reporting results, gather the required input and hand off to the shared **
 
 Read `plugin/skills/muggle-pr-visual-walkthrough/e2e-report-assembly.md` for the full assembly guide.
 
-#### 10b: Detect the PR, then apply the `postPRVisualWalkthrough` gate
+#### 10b: Post the walkthrough
```
stan4git: Do the prefgate check first, then call the skill to post

plugin/skills/muggle-test/SKILL.md
stan4git: Reworded line 76 to drop the ambiguous "test my changes" hint. New text explicitly distinguishes the use cases: Mode C = repo-wide sweep across **every** open PR regardless of author (not 'test my branch', not 'test my stack'); for a single branch or a stack of your own PRs use Mode A (local) or Mode B (preview/staging) on each branch.

plugin/skills/muggle-test-feature-local/SKILL.md
stan4git: Reordered. Gate fires first (step 1), PR existence + autoCreatePR check next (steps 2-3), and only then assemble the E2eReport + invoke the walkthrough skill (steps 4-5). Same restructure applied in muggle-test Step 9. If the user skips at the gate, the assembly work is skipped too — no wasted effort.

PR #142, opened 2026-05-12
chore(skills): gate-first ordering for walkthrough post

stan4git:
## Summary

Replaces the verbose Case A / Case B narration in `muggle-test` Step 9 and `muggle-test-feature-local` Step 10 with a single 5-step gate-first list. (Picked up from the closed #141, which had this cleanup but is no longer mergeable after master diverged.)

**New shape — same in both files:**

1. Fire `postPRVisualWalkthrough` — on skip, stop.
2. Look up the PR.
3. If no PR, fire `autoCreatePR` — on skip, stop.
4. Assemble the `E2eReport`.
5. Hand off to `muggle-pr-visual-walkthrough` Mode A.

**Why this is better:**

- **Gate fires first** — skipping at the gate also skips the assembly work. The old order assembled the report speculatively and then discarded it on skip.
- **No more 9a/9b/9c (or 10a/10b/10c) sub-sections** — the Case A/B logic now lives in the gate files (`postPRVisualWalkthrough.md`, `autoCreatePR.md`), not duplicated in every caller's procedure.
- **Diff is net `-32` lines.**

## Test plan

- [ ] Confirm no other docs reference `Step 9a`/`9b`/`9c` or `Step 10a`/`10b`/`10c` (grep clean in `plugin/`).
- [ ] Confirm the Preferences tables in both files point `autoCreatePR` at the parent step (`9` / `10`) since the sub-step `9b`/`10b` no longer exists.
- [ ] Spot-check that the Step 9/10 procedure still reads correctly end-to-end.

### DIFF
--- a/plugin/skills/muggle-test/SKILL.md
+++ b/plugin/skills/muggle-test/SKILL.md
@@ -48,7 +48,7 @@ Gates run per `preference-gates/README.md`.
 | `autoPublishLocalResults` | 7A | Upload local results to Muggle Test cloud after run |
 | `showElectronBrowser` | 7A | Show the Electron browser window during local test execution (vs. run headless) |
 | `postPRVisualWalkthrough` | 9 | Post visual walkthrough to PR after results are available |
-| `autoCreatePR` | 9 | Auto-create the PR when posting the walkthrough has no PR to target |
+| `autoCreatePR` | 9 (if no PR) | Auto-create the PR when posting the walkthrough has no PR to target |
 
 ## Step 1: Confirm Scope of Work (Always First)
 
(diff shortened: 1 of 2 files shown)


plugin/skills/muggle-test/SKILL.md
```diff
@@ -48,7 +48,7 @@ Gates run per `preference-gates/README.md`.
 | `autoPublishLocalResults` | 7A | Upload local results to Muggle Test cloud after run |
 | `showElectronBrowser` | 7A | Show the Electron browser window during local test execution (vs. run headless) |
 | `postPRVisualWalkthrough` | 9 | Post visual walkthrough to PR after results are available |
-| `autoCreatePR` | 9b | Auto-create the PR when posting the walkthrough has no PR to target |
+| `autoCreatePR` | 9 | Auto-create the PR when posting the walkthrough has no PR to target |
```
stan4git: why two 9s?

stan4git: Differentiated the step labels: postPRVisualWalkthrough stays at "9"; autoCreatePR is now "9 (if no PR)" (same in muggle-test-feature-local: "10" / "10 (if no PR)"). Same step number — but the label now says when each one fires, so the two rows aren't visually redundant. Commit 216039b.

PR #165, opened 2026-05-20
fix(replay): localUrl swap covers subdomains + forces first navigate (#163)

### LINKED ISSUE #163: muggle-local-execute-replay: localUrl swap misses step-level navigate URLs (replay still hits cloud)
## Summary

When running `muggle-local-execute-replay`, the `localUrl` parameter only rewrites occurrences of the **exact** `testScript.url` string in the action script. Step-level `operation.url` values that point to *related* hostnames (e.g. the auth subdomain a recorded session redirected through) are **not** rewritten, so the first `navigate` step sends the replay browser straight to the original cloud environment, defeating the entire purpose of \"replay against localhost.\"

## Reproduction

1. Pick any test case whose recorded action script's first step is a `navigate` to a non-product URL (e.g. an Auth0 redirect like `https://login.staging.muggle-ai.com/u/login?state=...`) — this is normal for any login-gated flow.

2. Invoke `muggle-local-execute-replay` with:

   ```jsonc
   {
     \"testScript\": {
       \"url\": \"https://staging.muggle-ai.com/muggleTestV0/dashboard\",  // product domain
       // ... rest of testScript ...
     },
     \"actionScript\": [
       {
         \"operation\": {
           \"action\": \"navigate\",
           \"url\": \"https://login.staging.muggle-ai.com/u/login?state=...\"  // *auth* domain
         }
         // ...
       }
       // ... rest of steps ...
     ],
     \"localUrl\": \"http://localhost:3999\"
   }
   ```

3. Observed: the Electron browser navigates to `https://login.staging.muggle-ai.com/u/login?state=...` (staging Auth0), **not** to `http://localhost:3999`. The replay continues operating against the *staging* environment until something diverges (selector mismatch, staging outage, etc.).

In a session today the navigate landed on this staging Auth0 error page (because staging Auth0 itself was flaky at the moment) and the replay died with \"Invalid element ID: 7\" on step #2:

> **staging-muggleai — Oops!, something went wrong. There could be a misconfiguration in the system or a service outage.**

The user expected `localUrl` to send the browser to `localhost:3999`.

## Root cause

`packages/mcps/src/mcp/local/services/execution-service.ts` → `rewriteActionScriptUrls`:

\`\`\`ts
function rewriteActionScriptUrls(params: {
  actionScript: unknown[];
  originalUrl?: string;
  localUrl: string;
}): unknown[] {
  const { actionScript, originalUrl, localUrl } = params;
  if (!originalUrl) return actionScript;

  // Deep clone and replace URLs
  const serialized = JSON.stringify(actionScript);
  const rewritten = serialized.replace(new RegExp(escapeRegex(originalUrl), \"g\"), localUrl);
  return JSON.parse(rewritten) as unknown[];
}
\`\`\`

It substitutes only the literal `originalUrl` (= `testScript.url`, the product/dashboard URL). The auth subdomain (`login.<product-host>`) is a different string and slips through. Any other hostname related to the original origin (CDN, websocket, etc.) is also missed.

## Proposed fix

Two parts; either alone is an improvement, but together they're the right contract:

### 1. Force the first `navigate` step's URL to be `localUrl`

The replay is, by definition, against `localUrl`. The recorded first navigate is implicitly \"start the user journey here\" — it has to start at the local origin, otherwise the local environment never gets touched.

\`\`\`ts
// inside rewriteActionScriptUrls, after the regex substitution:
const result = JSON.parse(rewritten) as Array<Record<string, unknown>>;
const firstNavigate = result.find((step) => {
  const op = (step as { operation?: { action?: string } }).operation;
  return op?.action === \"navigate\";
});
if (firstNavigate) {
  const op = firstNavigate.operation as Record<string, unknown>;
  op.url = localUrl;
}
return result;
\`\`\`

### 2. Substitute *any* URL whose hostname equals or is a subdomain of `originalUrl`'s hostname

So `https://login.staging.muggle-ai.com/...` gets rewritten when `originalUrl`'s host is `staging.muggle-ai.com`. This catches auth subdomains, websocket subdomains, CDN subdomains of the same product origin — but not unrelated hosts like `accounts.google.com`.

\`\`\`ts
function rewriteActionScriptUrls(params: { ... }): unknown[] {
  const { actionScript, originalUrl, localUrl } = params;
  if (!originalUrl) return actionScript;

  const originalHost = new URL(originalUrl).hostname;
  // hostname matches if it equals `originalHost` OR ends with `.${originalHost}`
  const hostPattern = new RegExp(
    \`https?://(?:[a-z0-9-]+\\.)*\${escapeRegex(originalHost)}(?=[/?#:]|$)\`,
    \"gi\",
  );

  const localOrigin = new URL(localUrl).origin;
  const serialized = JSON.stringify(actionScript);
  const rewritten = serialized.replace(hostPattern, localOrigin);

  // ... then force first-navigate to localUrl as above ...
}
\`\`\`

## Test plan

- [ ] Unit test: action script with step-level `operation.url` on `login.staging.muggle-ai.com` and `testScript.url` on `staging.muggle-ai.com/muggleTestV0/dashboard` + `localUrl=http://localhost:3999` → rewritten step URL starts with `http://localhost:3999`.
- [ ] Unit test: action script with `accounts.google.com` in a step URL → **not** rewritten (different host).
- [ ] Unit test: action script with no navigate step → unchanged.
- [ ] Smoke test: replay an existing login-gated test case against localhost; first navigate lands on `localhost:3999`.

## Impact

- Any login-gated replay against a localUrl is currently silently running against the original cloud env. This is a) misleading (claims to be \"local\") and b) brittle (cloud env outages bleed into local replay).
- Fixing this unlocks the meaningful local-replay workflow (and would have unblocked the PR 303 walkthrough on muggle-ai-ui that surfaced this in the first place).

## Evidence

Replay run ID `8a75d444-64e7-4014-8d31-76efeceff2c8` (Electron exit 27, 43s). Step #1 navigated to `https://login.staging.muggle-ai.com/u/login?state=...`; step #2 captured a screenshot of the staging Auth0 \"Oops!\" error page, not localhost.

Found while running `/muggle-test` on muggle-ai-ui PR #303 with `localUrl=http://localhost:3999`.

stan4git:
## Summary

Closes #163. Two-part fix inside `rewriteActionScriptUrls`, extracted from `execution-service.ts` into a dedicated `replay-url-rewrite.ts` so it's testable in isolation.

1. **Hostname-aware substitution.** Rewrite any URL whose hostname equals or is a subdomain of `originalUrl`'s hostname to `localUrl`'s origin — `staging.muggle-ai.com` *and* `login.staging.muggle-ai.com` both rewrite, while unrelated hosts like `accounts.google.com` are left alone. Falls back to the previous literal-string substitution when `originalUrl` doesn't parse as a URL.

2. **First-navigate override.** After the hostname pass, the first `navigate` step's `operation.url` is forced to `localUrl`. The recorded value is typically a stateful auth redirect (`login.<host>/u/login?state=...`) that's invalid in the local env; the local app's own auth redirect chain has to start from `localUrl`.

## Why this matters

Before: `muggle-local-execute-replay` claimed to run against `localUrl` but actually navigated to the original cloud host in step #1 (the auth subdomain), silently testing the cloud env. PR 303's walkthrough on `muggle-ai-ui` hit this — the replay's first step landed on staging Auth0's `Oops, something went wrong` page instead of `localhost:3999`. See #163 for the full reproduction.

After: `localUrl` is honored end-to-end. The browser starts at the local origin, the local app drives its own auth redirect chain.

## Files

- `packages/mcps/src/mcp/local/services/replay-url-rewrite.ts` *(new)* — the rewrite logic + helpers
- `packages/mcps/src/mcp/local/services/execution-service.ts` — drop the inlined function, import from the new module
- `packages/mcps/src/test/replay-url-rewrite.test.ts` *(new)* — 8 vitest cases

## Test plan

- [x] **Unit:** `pnpm exec vitest run src/test/replay-url-rewrite.test.ts` — 8 tests passing
  - returns input unchanged when `originalUrl` is missing
  - first navigate forced to `localUrl` regardless of recorded value
  - subdomain hostnames (`login.staging.muggle-ai.com`) rewritten everywhere they appear
  - unrelated hosts (`accounts.google.com`) not rewritten
  - scripts with no navigate step process correctly
  - only the *first* navigate is forced; later navigates use the hostname-rewritten URL
  - non-default `localUrl` ports preserved
  - non-parseable `originalUrl` falls back to literal substitution
- [x] `pnpm run lint:check`
- [x] `pnpm run typecheck`
- [x] `pnpm test` — full suite
- [x] `pnpm run build`
- [x] `pnpm run verify:plugin`
- [x] `pnpm run verify:contracts`
- [x] `pnpm run verify:electron-release-checksums`
- [ ] Smoke test post-release: invoke `muggle-local-execute-replay` against `muggle-ai-ui` PR 303 with `localUrl=http://localhost:3999`; first step's screenshot lands on `localhost:3999`, not staging Auth0.

### DIFF
--- a/packages/mcps/src/mcp/local/services/replay-url-rewrite.ts
+++ b/packages/mcps/src/mcp/local/services/replay-url-rewrite.ts
@@ -1,34 +1,10 @@
-/**
- * URL rewriting for action-script replay against a local dev URL.
- *
- * Cloud-recorded action scripts encode the original environment's URLs in
- * both the top-level testScript metadata and inside individual step
- * operations (especially the first `navigate`, which is usually a stateful
- * auth redirect like `login.<host>/u/login?state=...`). When replaying
- * locally we want those URLs pointed at the local origin so the journey
- * actually exercises the local app.
- */
+/** Point cloud-recorded action script URLs at a local dev URL for replay. */
 
 /**
- * Rewrite URLs in an action script to use `localUrl` instead of the cloud
- * URL the script was recorded against.
- *
- * Substitution covers two cases:
- *   1. Any URL whose hostname equals or is a subdomain of `originalUrl`'s
- *      hostname is rewritten to `localUrl`'s origin (so e.g.
- *      `login.staging.muggle-ai.com/...` rewrites alongside
- *      `staging.muggle-ai.com/...` when the original is `staging.muggle-ai.com`).
- *      Unrelated hosts like `accounts.google.com` are left alone.
- *   2. The first `navigate` step's URL is forced to `localUrl` regardless of
- *      what was recorded. The recorded URL is often a stateful auth redirect
- *      that's invalid in the local env; starting the journey at `localUrl`
- *      lets the local app's own auth redirect chain take over.
- *
- * @param params - Rewrite parameters.
- * @param params.actionScript - Original action script steps.
- * @param params.originalUrl - Original cloud URL the script was recorded against.
- * @param params.localUrl - Local URL to redirect to.
- * @returns Action script with rewritten URLs.
+ * Rewrite URLs in an action script for local replay: any URL whose host
+ * equals or is a subdomain of `originalUrl`'s host swaps to `localUrl`'s
+ * origin, and the first `navigate` step is forced to `localUrl` (the
+ * recorded value is usually a stateful auth redirect that can't replay).
  */
 export function rewriteActionScriptUrls(params: {
   actionScript: unknown[];
@@ -65,9 +41,6 @@ export function rewriteActionScriptUrls(params: {
   return result;
 }
 
-/**
- * Force the first `navigate` step's URL to `localUrl`. Mutates `steps`.
- */
 function forceFirstNavigateUrl(params: { steps: unknown[]; localUrl: string }): void {
   for (const step of params.steps) {
     if (!isPlainObject(step)) continue;
--- a/packages/mcps/src/mcp/local/services/replay-url-rewrite.ts
+++ b/packages/mcps/src/mcp/local/services/replay-url-rewrite.ts
@@ -1,0 +1,88 @@
+/**
+ * URL rewriting for action-script replay against a local dev URL.
+ *
+ * Cloud-recorded action scripts encode the original environment's URLs in
+ * both the top-level testScript metadata and inside individual step
+ * operations (especially the first `navigate`, which is usually a stateful
+ * auth redirect like `login.<host>/u/login?state=...`). When replaying
+ * locally we want those URLs pointed at the local origin so the journey
+ * actually exercises the local app.
+ */
+
+/**
+ * Rewrite URLs in an action script to use `localUrl` instead of the cloud
+ * URL the script was recorded against.
+ *
+ * Substitution covers two cases:
+ *   1. Any URL whose hostname equals or is a subdomain of `originalUrl`'s
+ *      hostname is rewritten to `localUrl`'s origin (so e.g.
+ *      `login.staging.muggle-ai.com/...` rewrites alongside
+ *      `staging.muggle-ai.com/...` when the original is `staging.muggle-ai.com`).
+ *      Unrelated hosts like `accounts.google.com` are left alone.
+ *   2. The first `navigate` step's URL is forced to `localUrl` regardless of
+ *      what was recorded. The recorded URL is often a stateful auth redirect
+ *      that's invalid in the local env; starting the journey at `localUrl`
+ *      lets the local app's own auth redirect chain take over.
+ *
+ * @param params - Rewrite parameters.
+ * @param params.actionScript - Original action script steps.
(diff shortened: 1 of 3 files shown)


packages/mcps/src/mcp/local/services/replay-url-rewrite.ts
```diff
@@ -0,0 +1,88 @@
+/**
+ * URL rewriting for action-script replay against a local dev URL.
```
stan4git: tight up comment

stan4git: Tightened in 0e130b6: file header is now a single line, function-level block dropped the enumerated list / @param tags, and the never-actually-needed JSDoc on the private `forceFirstNavigateUrl` is gone. Net –31 lines of prose, 8 tests still green.

PR #177, opened 2026-05-22
feat(local-exec): cross-worktree single-flight lock + PR-worktree skill flow

stan4git:
## Summary
- Adds a file-based single-flight lock around local Electron execution so concurrent muggle-test runs from different worktrees serialize. Same `cwd` = reentrant (intra-change parallel still allowed); different `cwd` = blocks until released.
- Updates the muggle-test skill so a PR URL materializes a dedicated worktree at `<repo>/.claude/worktrees/<branch>` instead of switching the user's main checkout. The worktree path is threaded through to a new `cwd` parameter on `muggle-local-execute-test-generation` and `muggle-local-execute-replay`, which is what keys the lock.

## Implementation notes
- Lock file at `~/.muggle-ai/local-execution.lock` with `{ holderCwd, holderPids: number[], acquiredAt }`. Sidecar `.modify` file atomically held during read-modify-write so two processes can't race on the state.
- Reentrant via the pid list: same-cwd callers append their pid; release removes one occurrence; file is deleted when the list empties.
- Stale detection: if every holder pid is dead (`process.kill(pid, 0)` raises ESRCH), the next caller breaks and takes over.
- `cwd` parameter is optional on both MCP tool schemas; tool-registry defaults to `process.cwd()` if omitted, so old skill callers still work.

## Test plan
- [x] 4 new unit tests in `packages/mcps/src/test/local-execution-lock.test.ts` — acquire/release, refcount reentrancy, different-cwd waits, stale-lock takeover.
- [x] Full mcps vitest suite green (130 tests).
- [x] `tsc --noEmit` clean.
- [ ] After release: run `/muggle-test https://github.com/.../pull/N` end-to-end and confirm the PR worktree is created and the lock log shows on the second invocation.

### DIFF
--- a/packages/mcps/src/mcp/local/services/execution-service.ts
+++ b/packages/mcps/src/mcp/local/services/execution-service.ts
@@ -23,15 +23,12 @@ import {
 import { getLogger } from "../../../shared/logger.js";
 import type { TestCaseDetails, TestScriptDetails } from "../contracts/project-schemas.js";
 import { getAuthService, getRunResultStorageService, getStorageService } from "./index.js";
+import { acquireLocalExecutionLock } from "./local-execution-lock.js";
 import { rewriteActionScriptUrls } from "./replay-url-rewrite.js";
 import type { ILocalExecutionContext } from "./run-result-storage-service.js";
 import type { RunResultStatus } from "./run-result-storage-service.js";
 
 const logger = getLogger();
-
-// ========================================
-// Types
-// ========================================
 
 /**
  * Internal execution process tracking.
@@ -81,16 +78,8 @@ export interface ILocalRunResult {
   errorMessage?: string;
 }
 
-// ========================================
-// Active Process Tracking
-// ========================================
-
 /** Map of active execution processes. */
 const activeProcesses: Map<string, IInternalExecutionProcess> = new Map();
-
-// ========================================
-// Auth Helpers
-// ========================================
 
 /**
  * Get the authenticated user ID.
@@ -199,10 +188,6 @@ function buildStudioAuthContent(): { accessToken: string; email: string; userId:
     userId: storedAuth.userId,
   };
 }
-
-// ========================================
-// Temp File Helpers
-// ========================================
 
 /**
  * Ensure temp directory exists.
@@ -600,10 +585,6 @@ async function executeElectronAppAsync(params: {
   });
 }
 
-// ========================================
-// Main Execution Functions
-// ========================================
-
 /**
  * Execute test script generation for a test case.
  *
@@ -617,6 +598,23 @@ async function executeElectronAppAsync(params: {
  * @returns Run result with generated script info.
  */
 export async function executeTestGeneration(params: {
+  testCase: TestCaseDetails;
+  localUrl: string;
+  cwd: string;
+  mutations?: string[];
+  timeoutMs?: number;
+  showUi?: boolean;
+  freshSession?: boolean;
+}): Promise<ILocalRunResult> {
+  const lockHandle = await acquireLocalExecutionLock({ cwd: params.cwd });
+  try {
+    return await runTestGenerationLocked(params);
+  } finally {
+    await lockHandle.release();
+  }
+}
+
+async function runTestGenerationLocked(params: {
   testCase: TestCaseDetails;
   localUrl: string;
   mutations?: string[];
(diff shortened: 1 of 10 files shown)


packages/mcps/src/mcp/local/services/execution-service.ts
```diff
@@ -617,6 +598,23 @@ async function executeElectronAppAsync(params: {
  * @returns Run result with generated script info.
  */
 export async function executeTestGeneration(params: {
+  testCase: TestCaseDetails;
+  localUrl: string;
+  cwd: string;
```
stan4git, line 604: why this is required while definition is optional?

packages/mcps/src/mcp/local/services/local-execution-lock.ts
```diff
@@ -0,0 +1,256 @@
...
+
+const LOCK_FILE_NAME = "local-execution.lock";
+const MODIFY_LOCK_SUFFIX = ".modify";
+const POLL_INTERVAL_MS = 2000;
+const WAIT_LOG_INTERVAL_MS = 10_000;
+const MODIFY_LOCK_RETRY_MS = 50;
+/** Stale modify-lock threshold. If the sidecar persists past this, assume crash. */
+const MODIFY_LOCK_STALE_MS = 5_000;
+
+interface ILockState {
```
stan4git: move out

packages/mcps/src/mcp/local/services/local-execution-lock.ts
```diff
@@ -0,0 +1,256 @@
...
+  try {
+    await fileHandle.writeFile(JSON.stringify({ pid: process.pid, ts: Date.now() }));
+    return await fn();
+  } finally {
+    await fileHandle.close().catch(() => undefined);
+    await fs.unlink(modifyPath).catch(() => undefined);
+  }
+}
+
+interface ITryAcquireOutcome {
```
stan4git: move out

packages/mcps/src/mcp/tools/local/tool-registry.ts
```diff
@@ -342,6 +330,7 @@ const executeTestGenerationTool: ILocalMcpTool = {
       const result = await executeTestGeneration({
         testCase: input.testCase,
         localUrl: input.localUrl,
+        cwd: input.cwd ?? process.cwd(),
```
stan4git: how this would not caused two locks? possibly

plugin/skills/muggle-test/SKILL.md
```diff
@@ -96,7 +96,14 @@ Analyze the changes to understand what's impacted. Two sources, picked by what t
 **PR URL** (user passed `github.com/<org>/<repo>/pull/<n>`):
 1. `gh pr diff <n> --repo <org>/<repo> --name-only` for the changed file list
 2. `gh pr diff <n> --repo <org>/<repo>` for the actual diff
-3. The repo lives at a sibling path (e.g. `C:\Users\stan4\Github\<repo>`) — `cd` into it and verify the PR branch is checked out before running tests; if not, ask the user to check it out (or offer to do it).
+3. Locate the repo at the sibling path (e.g. `C:\Users\stan4\Github\<repo>`).
```
stan4git: move this out and reference. NO need for this details here

packages/mcps/src/mcp/local/services/local-execution-lock.ts
stan4git: Moved to `local-execution-lock-constants.ts` so the logic file stays logic-only. Pushed in f8f34ef.

packages/mcps/src/mcp/local/services/local-execution-lock.ts
stan4git: Moved `ILockState`, `ILocalExecutionLockHandle`, and `ITryAcquireOutcome` to `local-execution-lock-types.ts`. Pushed in f8f34ef.

packages/mcps/src/mcp/tools/local/tool-registry.ts
stan4git: Real concern — exactly the scenario where the fallback could split one logical change into two competing holders (skill A passes `cwd=/worktrees/foo`, skill B omits cwd → defaults to MCP startup dir, which may or may not be the same path; even when they resolve to the same logical worktree, casing and trailing-slash differences could make them mis-key). Removed the fallback entirely. Schema now requires `cwd`, so the skill must pass it explicitly. Pushed in f8f34ef.

plugin/skills/muggle-test/SKILL.md
stan4git: Moved the worktree-creation details to `_shared/pr-branch-worktree.md` (matches the existing `_shared/failure-mode-handling.md` pattern). SKILL.md now just says "materialize per [`_shared/pr-branch-worktree.md`](...)" and notes the cwd handoff. Pushed in f8f34ef.

packages/mcps/src/mcp/local/services/execution-service.ts
stan4git: Aligned — `cwd` is now REQUIRED in both `Execute*InputSchema`s and the tool-registry no longer falls back to `process.cwd()`. The skill is the source of truth (always passes the worktree path), and the function signature matches the schema. Pushed in f8f34ef.

PR #180, opened 2026-05-22
feat(muggle-test-prepare): comprehensive smoke test + clean-start default

stan4git:
## Summary

- Add Step 7 to `muggle-test-prepare`: a mandatory HTTP + body-sniff + log-tail comprehensive check for **every** service in the tracking file (skill-managed and `external: true` alike) before the final readiness report.
- Add a diagnose-and-fix loop with **Clean Restart** (kill + fresh install + restart + re-verify) as the default option when a service fails the smoke test. Capped at 3 iterations per service.
- Make Step 5.5 (`npm install` when `node_modules/` is missing or stale) automatic with a notification instead of an `AskUserQuestion`. The only opt-out is aborting the skill.
- Extend `_shared/dev-server-readiness.md` with a Body Sniff Patterns section covering Next.js, Vite, generic Node, and raw stack frames, plus OS-agnostic regex examples for bash and PowerShell.
- Rename old Step 7 (Report Readiness) → Step 8 (Final Readiness Report), surfacing skipped services explicitly.
- Update the Integration Contract: `muggle-test-feature-local`, `muggle-do`, and local-mode `muggle-test` always invoke prepare; never bypass.

## Why

A dev server that compiled with type errors still binds to its port and returns `200 OK` — the response body is a webpack/vite error overlay. Port-listening alone treats this as healthy, tests then run against the overlay, and produce nonsense failures that look like product defects. The body sniff catches that whole class.

User report that triggered this: "I am still having broken UI (lint, build, missing deps issues), or required service not properly running" *after* `/muggle-test-prepare` claimed everything was ready.

## Test plan

- [ ] Invoke `/muggle-test-prepare` against a healthy local stack — confirm Step 7 passes for each service and Step 8 renders the readiness table.
- [ ] Manually break a service (e.g. introduce a TS error in a Next.js app, restart it) — confirm Step 7 catches it via the body sniff, the diagnose-and-fix loop fires, and Clean Restart is offered first.
- [ ] Stale `node_modules/` (delete it) — confirm Step 5.5 runs `npm install` automatically without an `AskUserQuestion`.
- [ ] User-started service (option 2 in Step 3) that's actually broken — confirm Step 7 catches it even though the skill doesn't own the log, and Options 1/2 (Clean Restart, Restart only) are NOT offered for `external: true` services.
- [ ] OS-agnostic body sniff works on both PowerShell and bash sample scripts in `_shared/dev-server-readiness.md`.

### DIFF
--- a/plugin/skills/muggle-test-prepare/SKILL.md
+++ b/plugin/skills/muggle-test-prepare/SKILL.md
@@ -62,21 +62,21 @@ Gates run per [`preference-gates/README.md`](../muggle-preferences/preference-ga
 
 ## Workflow
 
-Run the stages in this order. Each row links to its detail file — read the file when you reach the stage; the summary here is not the full instruction.
+Run the stages in this order. The sequence number is display-only — it lives only in this table for at-a-glance ordering; detail files and cross-references use slugs. Each row links to its detail file; read the file when you reach the stage.
 
-| Stage | Summary |
-|:------|:--------|
-| [rebase-check](./steps/rebase-check.md) | Rebase onto default branch (gated) |
-| [scope](./steps/scope.md) | Frontend / backend / full stack |
-| [viability-check](./steps/viability-check.md) | Exclude services that can't run locally |
-| [identify-services](./steps/identify-services.md) | Pick required services + startup mode |
-| [check-running](./steps/check-running.md) | Detect what's already listening |
-| [env-file](./steps/env-file.md) | Env file present + correct |
-| [start-commands](./steps/start-commands.md) | Determine per-service start command |
-| [fresh-install](./steps/fresh-install.md) | Auto-install deps if missing/stale |
-| [start-services](./steps/start-services.md) | Launch + two-stage readiness |
-| [smoke-test](./steps/smoke-test.md) | HTTP + body sniff + log tail; clean-restart on fail |
-| [readiness-report](./steps/readiness-report.md) | Final ready table |
+| # | Stage | Summary |
+|:--|:------|:--------|
+| 1 | [rebase-check](./steps/rebase-check.md) | Rebase onto default branch (gated) |
+| 2 | [scope](./steps/scope.md) | Frontend / backend / full stack |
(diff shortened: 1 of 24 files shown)


plugin/skills/muggle-test-prepare/SKILL.md
```diff
@@ -223,26 +223,26 @@ Use `AskUserQuestion`:
 
 If the user needs edits, collect corrections and re-present.
 
-### Step 5.5: Fresh-Worktree Install Probe
+### Step 5.5: Fresh Install (clean-start default)
```
stan4git: I feel like you can refactor this file so each step's detail instruction can be loaded from another file when it applies.

stan4git: Done in 7807c88. SKILL.md is now a scaffold (workflow index, Privacy Boundary, PID Tracking, Cleanup, Integration Contract, Guardrails) and each step's detail loads from `steps/step-N-<slug>.md`. Net same total content — just shifted to scaffold-here-content-there. 11 new step files, SKILL.md down from 369 to 85 lines.

plugin/skills/muggle-test-prepare/SKILL.md
```diff
@@ -43,266 +43,52 @@ All launched processes are tracked in `/tmp/muggle-test-prepare.json`:
...
-```
-
-If you launched the services, also show:
-```
-Logs: /tmp/muggle-prepare-*.log
-Cleanup: say "stop services" or re-invoke this skill.
-```
+Run the stages in this order. Each row links to its detail file — read the file when you reach the stage; the summary here is not the full instruction.
+
+| Stage | Summary |
```
stan4git: I would suggest to keep the stage seq number here for ordering perpose. not inside the sub file

stan4git: Done in a3ecedf. Sequence numbers restored as a display-only `#` column in the workflow table; step files and cross-references stay slug-only, so adding or reordering a stage means renumbering this one table and nothing else.

PR #181, opened 2026-05-22
feat(local-exec): preserve Electron per-step screenshots across all outcomes

stan4git:
## Summary
Failed local runs currently lose their per-step screenshots — on `exitCode !== 0`, `execution-service.ts` only writes stdout.log/stderr.log; the Electron app's `user_data/runtime/<launchTs>/screenshot/` contents are left to Electron's own (apparently aggressive) cleanup. The MCP server never copies them into the runId-keyed `sessions/<runId>/` dir.

Symptom: when a generation run dies with `goal_not_achievable` (which is the most common, most interesting failure mode the agent runs into), there's no way to see what the agent actually saw. That makes failure-mode classification per [`_shared/failure-mode-handling.md` section C](https://github.com/multiplex-ai/muggle-ai-works/blob/master/plugin/skills/_shared/failure-mode-handling.md#c-post-regen-failure-used-by-all-four-skills-with-generation) effectively impossible to do well — the `agent-course` bucket recommends invoking `/muggle-feedback`, but the user has no evidence to base feedback on.

## Implementation
Adds `preserveElectronRuntimeArtifacts()` that copies the Electron runtime dir into `sessions/<runId>/electron-runtime/` regardless of outcome. Two-tier strategy:

1. **Marker file** at `<tempDir>/<runId>_runtime_dir.txt` containing the absolute runtime-dir path. Written by Electron on startup — companion change in `muggle-ai-teaching-service` (separate PR). Gives reliable `runId` ↔ runtime correlation.
2. **mtime scan fallback**: scan `<electronCwd>/user_data/runtime/` for the most recent subdir with `mtime >= spawnAtMs`. Best-effort for older Electron builds that don't yet write the marker — preservation works even before the Electron-side companion ships.

Best-effort throughout — preservation failures log a warning and never break the run. `IElectronExecutionResult` now carries `electronCwd` so the caller can hand it to the helper without re-resolving the binary path.

## Rollout
- This PR can ship before the Electron-side marker change — the mtime fallback covers the gap. Once the Electron marker lands, correlation becomes deterministic instead of heuristic.
- No existing tests broken (130 → 126 because lock-related tests live on a different branch; same baseline as master).

## Test plan
- [x] `tsc --noEmit` clean
- [x] `vitest run` — 126/126 pass (no new tests added; the helper is best-effort and tested via the next E2E run)
- [ ] After merge + release: run `/muggle-test` on any PR, force a failure, confirm `sessions/<runId>/electron-runtime/screenshot/*.jpg` exist
- [ ] Verify with the Electron-side companion PR landed: marker file present → no mtime scan needed

### DIFF
--- a/packages/mcps/src/mcp/local/services/execution-service.ts
+++ b/packages/mcps/src/mcp/local/services/execution-service.ts
@@ -61,6 +61,8 @@ interface IElectronExecutionResult {
   stdout: string;
   /** Captured stderr. */
   stderr: string;
+  /** Electron's spawn cwd (path.dirname of the electron binary). Needed to locate user_data/runtime/ for artifact preservation. */
+  electronCwd: string;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


stan4git:
Rebased onto master to resolve conflict with #182 (which merged first). Conflict was purely whitespace — the two PRs compose cleanly:

- **#182** (merged): persists `stdout.log`, `stderr.log`, and (best-effort) `action-script.json` from `gen_*.json` on failure. Sets `artifactsDir` on the run result.
- **This PR (#181)**: persists the per-step `*.jpg` screenshots from Electron`s `user_data/runtime/<launchTs>/screenshot/` into `sessions/<runId>/electron-runtime/` regardless of outcome. Pairs with the merged Electron-side marker (muggle-ai-teaching-service#217).

After both land, a failed local run leaves: `stdout.log`, `stderr.log`, `action-script.json` (if produced), and `electron-runtime/screenshot/*.jpg` (if Electron created them before the crash). That covers every signal `_shared/failure-mode-handling.md` section C reads from.

Force-pushed `96881b8`; CI re-running.

PR #184, opened 2026-05-22
feat(mcps): consume backend-paginated test-runs/summary endpoint

stan4git:
> **DRAFT — blocked on [muggle-ai-prompt-service #530](https://github.com/multiplex-ai/muggle-ai-prompt-service/pull/530) being merged AND deployed.** Merging this MCP PR before the backend ships will 404 every call to `muggle-remote-project-test-runs-summary-get`.

## Summary

Switches `muggle-remote-project-test-runs-summary-get` from MCP-side pagination over the array endpoint to direct consumption of the new paginated endpoint added in prompt-service #530:

```
GET /v1/protected/muggle-test/projects/:id/test-runs/summary/paginated
    ?page=1&pageSize=20&sortBy=lastRunAt&sortOrder=desc
```

Each MCP page round-trip now pays for only one slice of JSON. The backend stops doing the N × 2 sequential workflow-run queries on every paginated MCP call, so this PR's value compounds the bigger the project gets.

### What changed

- **`mapToUpstream`** appends `?page=&pageSize=&sortBy=&sortOrder=` and points at the new `/paginated` path.
- **`mapFromUpstream`** simplifies to envelope unwrap + per-row slim. Sort, slice, and `totalCount` / `totalPages` / `hasMore` math all move to the backend.
- **Schema** (`ProjectTestRunsSummaryInputSchema`) already carries `page` / `pageSize` / `sortBy` / `sortOrder` from #183. No client-visible schema change.
- **Tool description** updated to advertise the new response shape (see below).
- **Tests** rewritten: 7 cases covering envelope passthrough, per-row slimming, missing-field omission, UNKNOWN status bucket, empty page, heavy payload size assertion, and graceful fallback when envelope fields are absent.

### What was dropped

`totals.byStatus` aggregate. The previous transform computed it because the MCP had visibility into the whole project's unpaginated payload. Now that the backend ships one page, a per-page byStatus would be misleading — an LLM seeing `FAILED: 18 / 2` on page 1 of 6 would draw a false 90%-failure conclusion. `totalCount` survives so the LLM still understands project scope. If a project-wide status histogram becomes an LLM requirement, it's a small additive change to the backend envelope.

### Response shape now

```json
{
  "page": 1,
  "pageSize": 20,
  "totalCount": 116,
  "totalPages": 6,
  "hasMore": true,
  "runs": [
    { "status": "TEST_REPLAY_FAILED", "testCaseId": "...", "testCaseTitle": "...",
      "useCaseId": "...", "useCaseTitle": "...", "lastRunAt": 1779409117068,
      "error": "Cannot read properties of undefined...", "latestWorkflowRunId": "..." }
  ]
}
```

## Test plan
- [x] `npm run test --filter @muggleai/mcp` — 137/137 passing (7 new transform tests + the 130 prior)
- [x] `npm run typecheck --filter @muggleai/mcp` — clean
- [x] `npm run build --filter @muggleai/mcp` — clean
- [ ] **Blocking**: prompt-service #530 merges + deploys
- [ ] Smoke: invoke `muggle-remote-project-test-runs-summary-get` with `projectId=5365f324-5c5f-4adf-b702-11f457b29b04` against the post-deploy backend and confirm the envelope shape lands

### DIFF
--- a/packages/mcps/src/mcp/tools/e2e/test-runs-summary-transform.ts
+++ b/packages/mcps/src/mcp/tools/e2e/test-runs-summary-transform.ts
@@ -4,6 +4,19 @@
  * keeps each row down to the fields an LLM actually reasons about so the
  * page slice doesn't carry duplicated useCase blocks and empty
  * studioAuthInfo nested objects across the wire.
+ *
+ * No input type lives here: the request shape is owned and validated by
+ * `ProjectTestRunsSummaryInputSchema` in `mcp/e2e/contracts`, and the tool
+ * handler in `tool-registry.ts` derives its parameter type from that schema
+ * via `z.infer`. Keeping a parallel TS interface here would create a second
+ * source of truth that could drift from the validator.
+ *
+ * No per-page aggregate (e.g. `byStatus`) is emitted: a status histogram
+ * computed over one page would describe only those entries and mislead the
+ * LLM about project-wide health (page 1 of 6 showing FAILED:18/PASSED:2
+ * reads like a 90% failure rate even when overall failures are 20 of ~116
+ * runs). If a project-wide histogram becomes an LLM requirement, the
+ * backend envelope should carry it.
  */
 
 import type { IUpstreamResponse } from "../../e2e/types.js";
--- a/packages/mcps/src/mcp/tools/e2e/test-runs-summary-transform.ts
+++ b/packages/mcps/src/mcp/tools/e2e/test-runs-summary-transform.ts
@@ -1,11 +1,9 @@
 /**
- * Slim + aggregate + paginate the upstream test-runs/summary payload.
- *
- * Upstream returns the full object graph (use case, test case, workflow run,
- * test script) per test-case row — ~150 lines × N test cases, which blows
- * past LLM token budgets on any non-trivial project. This transform keeps
- * only the fields an LLM needs to reason about run state and paginates the
- * result MCP-side (upstream does not support page params on this endpoint).
+ * Slim the per-row payload returned by the paginated test-runs/summary
+ * endpoint. The backend now handles sort + slice + envelope; this transform
+ * keeps each row down to the fields an LLM actually reasons about so the
+ * page slice doesn't carry duplicated useCase blocks and empty
+ * studioAuthInfo nested objects across the wire.
  */
 
 import type { IUpstreamResponse } from "../../e2e/types.js";
@@ -19,6 +17,15 @@ interface IRawSummaryEntry {
   error?: string;
 }
 
+interface IRawEnvelope {
+  data?: IRawSummaryEntry[];
+  page?: number;
+  pageSize?: number;
+  totalCount?: number;
+  totalPages?: number;
+  hasMore?: boolean;
+}
+
 interface ISlimSummaryEntry {
   status: string;
   testCaseId?: string;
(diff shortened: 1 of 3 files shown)


packages/mcps/src/mcp/tools/e2e/test-runs-summary-transform.ts
```diff
@@ -30,20 +37,10 @@ interface ISlimSummaryEntry {
...
-  page: number;
-  pageSize: number;
-  sortBy: "lastRunAt" | "status" | "testCaseTitle";
-  sortOrder: "asc" | "desc";
-}
-
 export interface ITestRunsSummaryOutput {
-  totals: {
-    total: number;
-    byStatus: Record<string, number>;
```
stan4git, line 43: why remove these?

packages/mcps/src/mcp/tools/e2e/test-runs-summary-transform.ts
```diff
@@ -30,20 +37,10 @@ interface ISlimSummaryEntry {
   latestWorkflowRunId?: string;
 }
 
-export interface ITestRunsSummaryInput {
```
stan4git, line 33: why delete?

stan4git: Backend now handles pagination, so the MCP no longer constructs an input object — `page`/`pageSize`/`sortBy`/`sortOrder` flow through as query params via the tool registry, and the shape is validated by `ProjectTestRunsSummaryInputSchema` in `contracts/index.ts`. The local interface had no remaining consumer after the rewrite.

packages/mcps/src/mcp/tools/e2e/test-runs-summary-transform.ts
stan4git: Dropped because per-page `byStatus` would mislead the LLM after the move to paginated fetching. With the backend returning one slice, a `byStatus` computed over that slice describes only those entries — e.g. `FAILED:18, PASSED:2` on page 1 of 6 reads like a 90% failure rate when it is really 20 of ~116 runs. Safer to omit than to invite that misreading.

packages/mcps/src/mcp/tools/e2e/test-runs-summary-transform.ts
stan4git: Addressed in 55d1e1d — pushed a header JSDoc to the file that explains the absence directly:

> No input type lives here: the request shape is owned and validated by `ProjectTestRunsSummaryInputSchema` in `mcp/e2e/contracts`, and the tool handler in `tool-registry.ts` derives its parameter type from that schema via `z.infer`. Keeping a parallel TS interface here would create a second source of truth that could drift from the validator.

packages/mcps/src/mcp/tools/e2e/test-runs-summary-transform.ts
stan4git: Addressed in 55d1e1d — same JSDoc update bakes in the rationale for the dropped `totals.byStatus`:

> No per-page aggregate (e.g. `byStatus`) is emitted: a status histogram computed over one page would describe only those entries and mislead the LLM about project-wide health (page 1 of 6 showing FAILED:18/PASSED:2 reads like a 90% failure rate even when overall failures are 20 of ~116 runs). If a project-wide histogram becomes an LLM requirement, the backend envelope should carry it.

PR #190, opened 2026-05-24
feat(pr-followup): seed E2E validation context at bootstrap

stan4git:
## Problem

A `muggle-pr-followup` watcher **bootstrapped from a PR URL** (`/muggle:muggle-pr-followup <pr-url>` — the documented entry) never runs pre-flight, so its `state.md` had no `localUrl` / `projectId` / validation strategy.

When a review landed and the address-reviews cycle reached **Stage 6 (E2E acceptance)**, `e2e-acceptance.md` Step 0 hard-halted:

> *"Missing any → pre-flight bug; escalate and halt; **do not ask the user**."*

So the cycle **neither ran E2E nor asked** — and because `open-prs/update.md` only posts a walkthrough *"if an E2E report exists,"* **nothing reached the PR** either. Only the per-comment replies fired. Forward-spawned watchers were unaffected (a forward `/muggle-do` run's pre-flight seeds the context into the shared session slot).

Secondary gap: `autoE2ETest: ask` is defined as "resolved by pre-flight Q13," which never runs in address-reviews mode.

## Fix

Gather the E2E validation context **once at bootstrap** (the user is present at launch), persist it to `state.md`, and have every later non-interactive tick reuse it. With context always present, Stage 6 runs each cycle → a report exists → the walkthrough posts. Replies already fired.

- **new `_shared/e2e-validation-context.md`** — the seed/consume contract: silent detection, the validation-subset questions (referencing pre-flight **Q4–Q8 / Q10** for canonical wording rather than forking them), and the persisted `state.md` fields.
- **`bootstrap.md`** — new **Step 6.5** runs the gather + persists; `--resume` backfills older sessions; reframed from "non-interactive" to "asks once, then autonomous."
- **`e2e-acceptance.md` Step 0** — reads the context **regardless of seeder** (pre-flight *or* bootstrap); in a loop there's no per-tick pre-flight, so the persisted `Validation` strategy is the standing decision (no `ask` re-resolution).
- **`address-reviews.md` 4d**, **`state-schemas.md`**, **`SKILL.md`**, **`CLAUDE.md`** — aligned with the new contract.

## Design decision

Pre-flight is **left untouched** (option b): the shared file *points at* pre-flight's questions as canonical wording rather than restructuring pre-flight's single-`AskUserQuestion` constraint. Zero blast radius on the working forward pipeline.

## Behavior guarantee

After a followup cycle addresses comments, it now (URL-bootstrapped path included): **runs E2E (or stands down per the strategy the user chose at bootstrap) → posts the result to the PR → replies per comment.**

### DIFF
--- a/plugin/skills/_shared/resolve-e2e-validation-context.md
+++ b/plugin/skills/_shared/resolve-e2e-validation-context.md
@@ -1,6 +1,10 @@
-# E2E Validation Context
+# Resolving the E2E Validation Context
 
-The context Stage 6 (E2E acceptance) needs to run unattended: validation strategy, local URL, backend status, Muggle Test project, test credentials, auth. A seeder resolves it once and writes it to `state.md` under a `## Pre-flight answers` block; the cycle reads it on every later run.
+**How-to procedure.** Followed by any seeder that prepares a session for unattended Stage 6 (E2E acceptance) runs.
+
+**Goal:** resolve — once, while the user is present — everything Stage 6 needs to run without further prompts (validation strategy, local URL, backend status, Muggle Test project, test credentials, auth), and persist it to `state.md` so every later non-interactive run reads it instead of asking.
+
+The sections below are the steps in order: reuse an existing context if one is on disk, else detect what's resolvable silently, ask the rest in one question, and write the result. The final [`## Persisted fields`](#persisted-fields) section doubles as the schema the cycle reads back.
 
 ## Reuse an existing context
 
--- a/plugin/skills/muggle-pr-followup/bootstrap.md
+++ b/plugin/skills/muggle-pr-followup/bootstrap.md
@@ -2,7 +2,7 @@
 
 The procedure for the **bootstrap mode** of `muggle-pr-followup` — invoked when a user dispatches the skill with a GitHub PR URL. Routing into this mode is documented in [`SKILL.md`](SKILL.md#routing).
 
-Bootstrap is **non-interactive**: it runs straight through, prompts the user for nothing, and ends with the first watcher dispatched. The first review the watcher sees triggers `/muggle-do`, where working-tree validation surfaces (via the existing E2E stage / muggle-test).
+Bootstrap asks **one** questionnaire — the E2E validation context the autonomous loop will reuse — then runs straight through to the first watcher dispatch. The user is present at launch, so this is the one place to gather it: every later watcher tick is non-interactive and reads the persisted context from `state.md`. Without it, an address-reviews cycle on a URL-bootstrapped watcher has no `localUrl`/`projectId` and Stage 6 hard-halts instead of running E2E.
 
 ## Turn preamble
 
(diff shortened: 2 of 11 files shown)


plugin/skills/muggle-pr-followup/bootstrap.md
```diff
@@ -2,7 +2,7 @@
 
 The procedure for the **bootstrap mode** of `muggle-pr-followup` — invoked when a user dispatches the skill with a GitHub PR URL. Routing into this mode is documented in [`SKILL.md`](SKILL.md#routing).
 
-Bootstrap is **non-interactive**: it runs straight through, prompts the user for nothing, and ends with the first watcher dispatched. The first review the watcher sees triggers `/muggle-do`, where working-tree validation surfaces (via the existing E2E stage / muggle-test).
+Bootstrap asks **one** questionnaire — the E2E validation context the loop will reuse — then runs through to the first watcher dispatch. The user is present at launch, so this is the only place to gather it; every later tick reads it from `state.md`. Without it, a URL-bootstrapped watcher has no `localUrl`/`projectId` and Stage 6 hard-halts instead of running E2E.
```
stan4git, line 5: If there is context, you should prompt user to reuse instead, and put a pref gate here to auto reuse if exists

plugin/skills/_shared/resolve-e2e-validation-context.md
stan4git, line 1: This file should not contains reverse dependencies to referencing docs

plugin/skills/muggle-pr-followup/bootstrap.md
stan4git: Done in 4a5411b. Added an `autoReuseValidationContext` gate: when a `## Pre-flight answers` block already exists for this working tree (current slot or most recent sibling session), the gather fires the gate before asking — `always` reuses it silently, `never` re-gathers, `ask` prompts reuse-vs-re-gather. Registered it in the `/muggle-preferences` configure picker ("Test setup" group). Host/project also still defer to the existing `autoSelectLocalHost`/`autoSelectProject` gates within a fresh gather.

plugin/skills/_shared/resolve-e2e-validation-context.md
stan4git: Done in 4a5411b. Dropped the Seeders/Consumer bullets and the caller-specific prose (forward-spawned vs URL-bootstrapped watcher) from the shared doc — it no longer names `bootstrap.md`/`e2e-acceptance.md` or any doc that references it. Remaining links are outbound only: pre-flight (canonical question wording), the preference gates, and `dev-server-readiness.md`. The seed/consume relationship is now declared by the callers, not back-referenced here.

stan4git, line 1: It is unclear to me if this is a schema or how-to doc?

Please use a better file name, and put a clear goal desc about what this does

stan4git: Done in ed45d69. It's a **how-to procedure** (the steps a seeder runs to resolve + persist the context), so I renamed it `_shared/e2e-validation-context.md` → `_shared/resolve-e2e-validation-context.md` — the verb signals it's an action, matching the other `_shared/` procedure docs. Updated all 6 referencing docs. The header now opens with an explicit "How-to procedure" label and a **Goal** line stating what it resolves and why (gather once while the user is present, so later non-interactive runs read it instead of asking); the `## Persisted fields` section is called out as the schema the cycle reads back.

PR #196, opened 2026-05-28
feat(internal): skill routing eval — entrances, harness, report

stan4git:
## Skill routing eval (internal tooling)

Internal-only tooling (not shipped in the published plugin) that defines the **triggering entrance** of every auto-trigger muggle skill and measures whether each skill's `description` routes the right queries to it — with all 14 skills competing in one live session. Lives under `internal/skill-routing-eval/`, alongside `internal/skill-gate-eval/`; pairs with the `optimize-descriptions` skill.

### What an "entrance" is
The surface where a skill becomes relevant and should be recommended: the intents/phrasings/contexts that route to it, and the sibling skills it must not steal from. `entrances.md` documents this for all 14 auto-trigger skills (the `m*` aliases and `disable-model-invocation` skills are out of scope — they fire only on explicit `/command` typing).

### How it's measured
`router_eval.py` runs each query through `claude -p "<query>" --max-turns 1` in this repo with all skills live. It reads the streamed events for the first tool call to detect the routed skill. `--max-turns 1` stops execution right after the routing decision, so **no skill body runs** — zero side effects. Every query runs 3× and the majority route is scored. Negative-class queries pass iff **no `muggle*` skill fires** (an appropriate non-muggle skill like `systematic-debugging` winning is correct).

### Results
- **Baseline: 79/82 = 96.3%** muggle-routing, 15/15 negatives clean, 3 genuine misses.
- **Final: see `reports/final.md`** after the three description fixes (#198/#199/#200).

### Files
`entrances.md` · `eval-set.json` (82 labeled queries) · `router_eval.py` · `analyze.py` · `reports/` (baseline + final, `optimization-log.md`, per-fix evidence under `reports/fixes/`).

The three description fixes ship as separate per-skill PRs based on this branch. The 11 clean skills route at 100% recall and are documented in `optimization-log.md` (no change needed).


stan4git:
I think this whole change should goes into internal skill folder.

stan4git:
Done — moved the whole thing into the internal skill folder.

It now lives at `internal/skill-routing-eval/` (alongside `internal/skill-gate-eval/`, and pairs with the `optimize-descriptions` skill), out of the published `plugin/skills/` surface entirely. Force-pushed this branch to the new path; nothing eval-related ships in the plugin anymore.

Restructured the stack to match:
- **This PR** now carries the complete internal package: entrances doc, harness, 82-query eval set, analyzer, and the full reports (baseline + final, optimization log, per-fix evidence).
- The **3 real description fixes** (`muggle-feedback` #198, `muggle-test` #199, `muggle-test-prepare` #200) are now plain `plugin/skills/<skill>/SKILL.md` changes based directly on this branch.
- The **11 clean-skill PRs** (#201–#211) are closed — after the move they were internal-doc no-ops; their verified entrances are recorded in the consolidated report here.

internal/skill-routing-eval/entrances.md
```diff
@@ -0,0 +1,68 @@
...
+**Engage when:** the user wants to **fix a broken Muggle install** — "muggle repair", "fix my muggle setup", "muggle MCP won't load, repair it", "reinstall/repair the muggle tooling".
+**Boundary vs `muggle-status`:** explicit fix intent. **Boundary vs `none`:** fixing the user's *app* build is not this.
+
+## muggle-upgrade
+**Engage when:** the user wants to **update Muggle to the latest version** — "muggle upgrade", "update muggle", "get the latest muggle".
+**Boundary vs `none`:** upgrading the user's npm deps / other tooling is not this. **Boundary vs `muggle-works-npm-release`:** consuming a new version ≠ publishing one.
+
+## muggle-works-npm-release
+**Engage when:** the user wants to **cut/publish a `@muggleai/works` npm release** — "release works", "cut a new works version", "publish @muggleai/works", "ship a patch release of works". Maintainer action on this repo.
+**Boundary vs `muggle-upgrade`:** publish (producer) vs update (consumer). **Boundary vs `none`:** releasing some *other* package is not this.
```
stan4git: This is more of a internal skill. Let's not have it here. move it properly.

internal/skill-routing-eval/eval-set.json
stan4git, line 1: Get 20-30 cases for each.

internal/skill-routing-eval/entrances.md
```diff
@@ -0,0 +1,68 @@
...
+**Boundary vs `muggle-test-import`:** no external source — operates on cases already in the project. **Boundary vs `muggle-test`:** project-wide script catch-up, not change-driven test execution.
+
+## muggle-test-prepare
+**Engage when:** the user needs **local dev servers / sibling services up before testing** — "are my services running", "spin up the dev servers", "get my local env ready for tests", "make sure localhost is up before we test". Also engaged *by* other muggle skills when ports aren't listening.
+**Boundary vs `muggle-test*`:** preparation/environment readiness, not the test run itself.
+
+## muggle-pr-visual-walkthrough
+**Engage when:** the user wants to **post/attach existing E2E results to a PR** — "post the test results to the PR", "add the visual walkthrough", "share the E2E screenshots on the PR", "put the pass/fail summary on the PR".
+**Boundary vs `muggle-test`:** purely the *publish-to-PR* step; assumes a run already happened.
+
```
stan4git, line 42: I would love muggle-pr-followup to be here. Can you make it a cmd/skill explicitly for user to use?

internal/skill-routing-eval/entrances.md
stan4git: Agreed — `muggle-works-npm-release` is maintainer-only (cutting the `@muggleai/works` npm release), not a customer routing target. Removed it from the entrances doc and the eval set. Relocating the actual skill out of the published plugin (`plugin/skills/muggle-works-npm-release` + the `mrelease` alias → `internal/skills/`) touches the plugin/marketplace/alias surface, so I'll do that as its own small PR rather than fold it into the eval — flag if you'd rather I skip it.

internal/skill-routing-eval/entrances.md
stan4git: Done — added an "Explicit-invocation skills" section to entrances.md documenting `muggle-pr-followup` (and `muggle-do`). It's already user-invocable as a command: `/muggle:muggle-pr-followup <pr-url>` to bootstrap a watcher (you just ran it), or `/loop 1m /muggle:muggle-pr-followup <slug> <n>` to poll. It carries `disable-model-invocation` so it never auto-fires from prose (correct for a watcher loop), which is why it's out of the auto-trigger eval. Want me to add a short alias (e.g. `/mfollow`) and/or surface it in the `muggle` router menu?

internal/skill-routing-eval/eval-set.json
stan4git: On it — expanding the eval set to ~20–25 concrete, varied queries per in-scope skill (13 skills after dropping works-release) plus a proportional negative set (~40), targeting ~330 cases. I'll keep them realistic and boundary-focused (the hard sibling collisions), and label each with its expected skill. One caveat I'll document: a single full pass over ~330 queries spawns ~1000 `claude -p` sessions and the muggle MCP intermittently disconnects mid-run (it corrupted two full passes already — the tail skills read 0% then re-probed 4/4 once reconnected), so I'll run it in per-skill chunks with the disconnect guard rather than one fragile sweep.

stan4git: Done — expanded to **364 cases** (was 78): 24–26 per in-scope skill + 45 negatives. Each skill's bank was authored by a dedicated pass grounded in the entrances contract, weighted toward the hard sibling boundaries (e.g. muggle-test vs feature-local, import vs regenerate-missing, status vs repair), and the negatives spread traps across every skill's vocab (npm/jest/vitest, import-a-library, upgrade-deps, k8s health, prettier config, product release notes, etc.). Committed in 04535a7.

Running the full set is ~1,092 `claude -p` sessions in one sweep, which the muggle MCP disconnect makes unreliable end-to-end, so I'll run it in per-skill chunks (with the disconnect guard) to refresh the report — say the word and I'll kick that off.

internal/skill-routing-eval/entrances.md
```diff
@@ -0,0 +1,73 @@
...
+## muggle-repair
+**Engage when:** the user wants to **fix a broken Muggle install** — "muggle repair", "fix my muggle setup", "muggle MCP won't load, repair it", "reinstall/repair the muggle tooling".
+**Boundary vs `muggle-status`:** explicit fix intent. **Boundary vs `none`:** fixing the user's *app* build is not this.
+
+## muggle-upgrade
+**Engage when:** the user wants to **update Muggle to the latest version** — "muggle upgrade", "update muggle", "get the latest muggle".
+**Boundary vs `none`:** upgrading the user's npm deps / other tooling is not this.
+
+## Explicit-invocation skills (command-only, out of auto-trigger scope)
```
stan4git: wait, muggle-pr-followup should also be optimized


muggle-works-npm-release should not show up here for optimization because it is jsut internal

stan4git: disable-model-invocation should not be "true" for pr follow up

stan4git: Done.

- **muggle-pr-followup is now model-invocable** — removed `disable-model-invocation: true` and rewrote its description around triggering intent ('watch my PR for review comments and address them', 'keep an eye on PR #123 and respond to reviews', 'babysit this PR's review thread'). Shipped in #218 (now merged into this branch).
- **Brought into the eval as a real entrance** — added its engage-when + boundaries (vs muggle-do / muggle-pr-visual-walkthrough / none) to entrances.md and **24 cases** (+3 near-miss negatives: merge-PR, summarize-reviews, ping-reviewers) to the eval set. Scope note updated: pr-followup in, works-release out.
- **muggle-works-npm-release**: agreed, internal-only — removed from the eval and relocated out of the published plugin in #217.
- **muggle-do** kept `disable-model-invocation` (autonomous implement-to-PR executor — shouldn't auto-fire); only the watcher was opened up. Say the word if you want it opened too.

pr-followup's triggering validation (router eval on the new cases) is queued behind the in-flight full run so it doesn't share the plugin cache; I'll post its recall + the refreshed report when that finishes.

PR #199, opened 2026-05-28
fix(muggle-test): own the pre-PR acceptance-gate intent

stan4git:
Stacked on #198.

### Miss
Baseline router eval: `validate my changes before I open the PR` routed to **`verification-before-completion` ×5** — a superpowers process skill owns "before creating PRs."

### Fix
The automated optimizer's full rewrite was **rejected on review** — it split the contested query across superpowers skills and regressed "regression test my work from today." Hand-authored instead: kept the original broad change-driven coverage and added an explicit clause that the pre-PR/merge acceptance gate means *run the acceptance suite (real browser, user flows), not just a completion checklist*, plus a `muggle-test-feature-local` boundary pointer.

### Evidence (targeted re-run, 9/9 @ 5×)
- `validate my changes before I open the PR` → `muggle-test` **5/5**
- `regression test my work from today` → still `muggle-test` (no regression)
- `muggle-test-feature-local` queries (checkout/signup/password-reset on localhost) → feature-local 5/5 (no collision)
- `post the e2e test results to the PR` → `muggle-pr-visual-walkthrough` (unaffected)
- `run the full test suite with npm test` → `none`

Run JSON: `plugin/skills/_routing-eval/reports/fixes/muggle-test.json`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #200, opened 2026-05-28
fix(muggle-test-prepare): claim port/service readiness checks before testing

stan4git:
Stacked on #199.

### Miss
Baseline router eval: `check if localhost:3000 and the api on 8080 are listening before testing` routed to **`none`** (fired only 2/5 on re-check) — the port-check phrasing read as a question Claude answers directly rather than a skill trigger.

### Fix (optimizer proposal, accepted on review)
Lead with environment readiness and explicitly claim confirming that *specific ports or localhost URLs are listening/up before testing* (with the exact example), plus a "this is environment readiness and service startup, not running the tests themselves" boundary so it doesn't poach `muggle-test`.

### Evidence (targeted re-run, 8/8)
- failing query → `muggle-test-prepare` 3/3
- 4 other prepare positives → pass
- `muggle-test` ("validate before PR", "test on staging") and `muggle-test-feature-local` ("checkout on localhost") → unaffected

Run JSON: `plugin/skills/_routing-eval/reports/fixes/muggle-test-prepare.json`.


stan4git:
⚠️ GitHub auto-marked this PR as *merged* during a base-branch force-push while relocating the eval into `internal/` (per review on #196). It did **not** merge to master. Replaced by #213 (same `muggle-test-prepare` fix, now based on #196). Please review #213 instead.

PR #229, opened 2026-05-30
refactor(dev-loop): single-source the dev loop (SoC: B)

stan4git:
First of the SoC consolidations (B). **Move the how-to into `_shared/`, reference it from the skills.**

## Problem
The local run loop was reimplemented 3–4× (`do/e2e-acceptance.md`, `muggle-test`, `muggle-test-feature-local`, `acceptance-tester`), with drifted copies of the MCP sequence, `timeoutMs`/exit-26 handling, `actionScript`-as-is, `freshSession`, result-reading, publish, screenshots.

## Change
- **New `_shared/dev-loop/`** — single source, broken into per-section files: `run.md`, `action-script.md`, `fresh-session.md`, `timeouts.md`, `failures.md`, `publish.md`. (Renamed from the initial `e2e-run.md` and split per review.)
- **Each caller** links the relevant file and keeps only its own glue.
- **Divergence fixed:** `muggle-test` now inherits the timeout/exit-26 guidance and zero-step publish fallback it lacked.

## Tests
`src/test/skills/` + `cleanup-skills` — 83 green; all links resolve.

## Sequence
B of B/C/D. Done: C #230, D #231. Plus #228.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #231, opened 2026-05-30
refactor(readiness): delegate to muggle-test-prepare (SoC: D)

stan4git:
SoC consolidation **D** (final of B/C/D): `muggle-test-prepare` owns service readiness + start, but two of its three mandated consumers bypassed it.

## Change
- **`muggle-do` pre-flight (items 3/4)** → invoke `muggle-test-prepare` instead of probing `dev-server-readiness.md` directly.
- **`muggle-do` `e2e-acceptance` Step 0.5** → readiness delegates to prepare (idempotent fast-exit; a mid-cycle down service is a pre-flight bug, not a prompt — preserves autonomy).
- **`muggle-test` local Step 7A** → adds the missing prepare invocation before URL selection (it previously did only an `lsof` port sniff).

`dev-server-readiness.md` remains the low-level probe that prepare uses internally; consumers stop calling it directly. One owner for readiness.

## Tests
`src/test/skills/` green; all `../muggle-test-prepare/SKILL.md` links resolve.

## Sequence — complete
- **B** #229 — E2E run-loop core
- **C** #230 — worktree path scheme + lifecycle ownership
- **D** (this) — readiness delegation

Plus #228 (cleanup trigger + tree guard, reworked to delegate). All four are the "move how-tos into `_shared/`, reference from skills" pass.

### DIFF
--- a/plugin/skills/do/e2e-acceptance.md
+++ b/plugin/skills/do/e2e-acceptance.md
@@ -58,7 +58,7 @@ For a `local-e2e` block, use `localUrl`, `projectId`, and the working-tree path 
 
 Before launching the local runner:
 
-1. **Dev-server + backend readiness** — invoke [`muggle-test-prepare`](../muggle-test-prepare/SKILL.md), the readiness/service-start owner (idempotent fast-exit when healthy; probes via `dev-server-readiness.md`). Halt on failure. In autonomous mode a service down mid-cycle is a pre-flight bug, not a prompt.
+1. **Dev-server + backend readiness** — invoke [`muggle-test-prepare`](../muggle-test-prepare/SKILL.md), the readiness/service-start owner (idempotent fast-exit when healthy; probes via `dev-server-readiness.md`). Halt on failure.
 2. **Auth** — `muggle-remote-auth-status` must be `authenticated`; else escalate.
 3. **Identity tenant/domain match** — if test credentials were marked `existing`, confirm the repo's configured identity tenant/domain matches the recorded tenant/domain. Mismatch → halt.
 
--- a/plugin/skills/do/e2e-acceptance.md
+++ b/plugin/skills/do/e2e-acceptance.md
@@ -58,7 +58,7 @@ For a `local-e2e` block, use `localUrl`, `projectId`, and the working-tree path 
 
 Before launching the local runner:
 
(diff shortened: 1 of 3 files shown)


plugin/skills/do/e2e-acceptance.md
```diff
@@ -58,7 +58,7 @@ For a `local-e2e` block, use `localUrl`, `projectId`, and the working-tree path
 
 Before launching the local runner:
 
-1. **Dev-server + backend readiness** — per [`../_shared/dev-server-readiness.md`](../_shared/dev-server-readiness.md) (port + compile log + backend health). Halt on any failure.
+1. **Dev-server + backend readiness** — invoke [`muggle-test-prepare`](../muggle-test-prepare/SKILL.md), the readiness/service-start owner (idempotent fast-exit when healthy; probes via `dev-server-readiness.md`). Halt on failure. In autonomous mode a service down mid-cycle is a pre-flight bug, not a prompt.
```
stan4git: what the fuck does the new phrase does help for?

plugin/skills/muggle-test/SKILL.md
```diff
@@ -235,6 +235,10 @@ If the user picks "Override one or more", let them flip the mode for any test ca
 
 ## Step 7A: Execute — Local Mode
 
+### Local environment readiness
```
stan4git, line 238: I feel this should belong to muggle-test-feature-local

plugin/skills/do/e2e-acceptance.md
stan4git: Cut in a45547f — the delegation + "Halt on failure" already carry it.

plugin/skills/muggle-test/SKILL.md
stan4git: Agreed — this is the real duplication: `muggle-test` §7A reimplements the local run that `muggle-test-feature-local` owns, which is exactly why the readiness step felt out of place here. Proper fix: the `muggle-test` local path delegates to `muggle-test-feature-local` so readiness + execution + the gates all live in one place. That is bigger than this readiness PR and needs `muggle-test-feature-local` to accept a pre-selected batch of test cases (today it is single-feature). Taking it as a focused follow-up — flagging here for the record.

PR #233, opened 2026-05-31
feat(muggle-do): opt-in autonomous rebase-conflict resolution (autoResolveConflicts)

stan4git:
Workstream **B** of the merged design ([architecture/2026-05-30-…-design.md](https://github.com/multiplex-ai/muggle-ai-brain/blob/master/architecture/2026-05-30-muggle-do-conflict-resolution-ci-autofix-and-browser-task-rename-design.md)). Independent of [#232](https://github.com/multiplex-ai/muggle-ai-works/pull/232) (rename); **C** (CI auto-fix) follows.

## What changes
- **New gate `autoResolveConflicts`** — default `never` = abort + escalate, **byte-for-byte today's behavior**; opt-in `always` = resolve autonomously. Auto-registered via the `preference-gates/` directory listing.
- **New `_shared/resolve-rebase-conflicts.md`** — the resolution body: capture the pre-rebase SHA, classify conflicts (mechanical: lockfiles/generated/formatting → deterministic; semantic: logic → reasoned 3-way), then a **mandatory verify gate** (build + unit + E2E per `autoE2ETest`). Pass → proceed; fail or unreconcilable semantic conflict → `git reset --hard <pre_rebase_sha>` + escalate.
- **`rebase-before-e2e.md`** branches conflict handling on the new gate (the old "never auto-resolve" is now the `never` path, unchanged).
- **`address-reviews.md` Step 0** — per-cycle rebase so the loop tracks master across cycles, not only at pre-flight.
- **`muggle-do:escalation`** event gains `kind: "rebase-conflict"`.

## Invariants
- Never pushes an auto-resolved rebase that hasn't passed the verify gate.
- The branch is always restorable to the captured pre-rebase SHA.
- `never` (default) is exactly today's behavior.

## Choices (per design defaults)
- **No** per-cycle semantic-conflict file cap — the verify-or-rollback gate is the safety net.
- Reuses the existing escalation event rather than minting a new one.

## Verification
- `verify-plugin-marketplace.mjs` → **passed** (built `dist/plugin`); `verify-compatibility-contracts.mjs` → **passed**. Gate file resolves via `ls preference-gates/*.md`; all cross-links resolve.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #234, opened 2026-05-31
feat(muggle-pr-followup): watch CI and auto-fix red checks

stan4git:
Workstream **C** of the merged design ([architecture/2026-05-30-…-design.md](https://github.com/multiplex-ai/muggle-ai-brain/blob/master/architecture/2026-05-30-muggle-do-conflict-resolution-ci-autofix-and-browser-task-rename-design.md)). Independent of [#232](https://github.com/multiplex-ai/muggle-ai-works/pull/232) (rename) and [#233](https://github.com/multiplex-ai/muggle-ai-works/pull/233) (conflict resolution).

## What changes
- **New recipe `_shared/github-cli-recipes/pr-checks.md`** — check-run rollup for the head SHA (`gh pr checks --json name,state,bucket,link`).
- **Watcher contract** now polls CI *after* reviews. **Reviews preempt CI** in a tick. With no new reviews: poll checks → pending = idle, green/none = idle, **red** dispatches `/muggle-do fix-ci` (PR url + slug + failing check names) while `ci_fix_attempts[sha] < 3` and `sha ∉ ci_escalated_shas`; otherwise idle.
- **New `last_seen` fields** `ci_fix_attempts` and `ci_escalated_shas`.
- **New `/muggle-do` route Fix-CI → `do/fix-ci.md`** (after address-reviews): map each red check to a local command (lint/format, typecheck, unit), **verify green before push**, commit `fix(ci): <check> — <what>`, **no PR replies** (the commit is the response). Out-of-scope checks (E2E-in-CI, infra, flaky) or a spent budget escalate (sha → `ci_escalated_shas`).
- **Telemetry**: `tick` gains `checks_red` + `dispatched_ci_fix`; `muggle-do:cycle` reuses with `ci_checks_*` fields and outcomes `ci-fixed` / `ci-escalated` (no new event).

## Choices
- **MAX 3** fix attempts per SHA (per your call). CI itself is the verify loop — each fix re-pushes, the next tick re-checks.
- Reviews always win a tick over CI; the dumb-pipe spirit holds (watcher detects + dispatches, never classifies/fixes).

## Verification
- `verify-plugin-marketplace.mjs` → **passed** (built `dist/plugin`); `verify-compatibility-contracts.mjs` → **passed**. New recipe added to the `github-cli-recipes` TOC; all cross-links resolve.

## Merge note
`muggle-do/SKILL.md` input-routing overlaps #232 (which renames route 3's target `muggle-do-task`→`muggle-browser-task`); this PR inserts a Fix-CI route and renumbers. Trivial merge-order resolution — apply both.

### DIFF
--- a/plugin/skills/_shared/github-cli-recipes.md
+++ b/plugin/skills/_shared/github-cli-recipes.md
@@ -10,6 +10,7 @@ Skills assume a working `gh auth status`. Auth errors surface verbatim from `gh`
 | :----- | :------- |
 | [`pr-metadata`](github-cli-recipes/pr-metadata.md) | Snapshot PR state, head SHA, branch — watcher + bootstrap. |
 | [`submitted-reviews`](github-cli-recipes/submitted-reviews.md) | Fetch reviews past a cursor — watcher's poll. |
+| [`pr-checks`](github-cli-recipes/pr-checks.md) | Check-run rollup for the head SHA — watcher's CI poll. |
 | [`line-comments-for-review`](github-cli-recipes/line-comments-for-review.md) | Pull a review's line comments — `/muggle-do` per-comment routing. |
 | [`unresolved-threads`](github-cli-recipes/unresolved-threads.md) | GraphQL fetch of unresolved comment threads — resolve-reminder. |
 | [`reply-line-comment`](github-cli-recipes/reply-line-comment.md) | POST a threaded reply on a line comment. |
--- a/plugin/skills/do/fix-ci.md
+++ b/plugin/skills/do/fix-ci.md
@@ -1,0 +1,66 @@
+# Fix-CI (watcher-dispatched)
+
+Resolve red CI on a PR's head — lint/format, typecheck, and failing unit tests — verifying green before re-push. Invoked by the [`muggle-pr-followup`](../muggle-pr-followup/contract.md) watcher (Step 5) when a tick sees failing checks and the fix budget for that SHA isn't spent. Like address-reviews, this is a dumb-pipe dispatch: the watcher detects, this stage fixes.
+
+## Turn preamble
+
+```
+**/muggle-do fix-ci** — fixing <count> red check(s) on <owner>/<repo>#<n>.
+```
+
+## Input
+
+`$ARGUMENTS` carries a `github.com/.../pull/<n>` URL, `slug=<slug>`, and the failing check names (no review ids). Parse all three.
+
+## Inputs from disk
+
+From `~/.muggle-ai/muggle-do/sessions/<slug>/`: `prs.json` (PR + local checkout / branch), `last_seen.json` (`ci_fix_attempts`, `ci_escalated_shas`, `pushed_shas`), `state.md` (worktree path, validation strategy).
+
+## Procedure
+
+### Step 1 — Re-attach
+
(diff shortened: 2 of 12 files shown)


plugin/skills/_shared/github-cli-recipes.md
stan4git, line 1: shouldn't this be in the folder of the same name?

plugin/skills/do/fix-ci.md
```diff
@@ -0,0 +1,66 @@
+# Fix-CI (watcher-dispatched)
```
stan4git, line 1: why reverse deps?

stan4git: This is how-to, doesn't belong here. maybe in shared?

plugin/skills/muggle-do/SKILL.md
stan4git, line 53: I feel this flow should not expand in do's skill file. move out into designated folder/files

plugin/skills/_shared/github-cli-recipes.md
stan4git: Kept as a sibling index. The repo convention is `X.md` beside `X/` — `pr-followup-helpers.md`, `telemetry-events.md`, and `do/open-prs.md` all sit next to their folders rather than inside them. Moving only this one would break that consistency. Happy to flip all three to in-folder indexes if you prefer that convention instead — just say the word.


plugin/skills/do/fix-ci.md
stan4git: Addressed in 3171739: dropped the reverse deps. fix-ci no longer links up to the watcher`s `contract.md` or restates its internal step ordering — the executor no longer references its caller. The only remaining mention is the functional respawn (`/loop ... muggle-pr-followup`), which re-arms the watcher after a push, identical to address-reviews Step 6.


stan4git: Addressed in 3171739: moved the "map red check -> local fix command" how-to into `_shared/ci-check-to-command.md`; fix-ci Step 2 now references it. The stage file keeps only orchestration (re-attach, verify, push, escalate).


plugin/skills/muggle-do/SKILL.md
stan4git: The fix-ci flow already lives in its own file, `do/fix-ci.md` — SKILL.md only gained two one-line routing pointers (Fix-CI, Post-merge cleanup), the same shape as the existing Address-reviews row. Routing is what SKILL.md is for (the `/muggle-do` entry router, consistent with `muggle-pr-followup/SKILL.md`). If you would rather the routing table itself move into a dedicated `do/input-routing.md` with SKILL.md holding just a pointer, I can do that — let me know.


stan4git: My concern is you are putting unrelated details into this highlevel guidence

stan4git: Addressed in 2ce24c9: moved the full input-routing table (and its watcher-dispatch specifics) out into a dedicated `do/input-routing.md`. SKILL.md now keeps only a one-paragraph mode overview plus a pointer, so the high-level guidance stays high-level. Thanks for the steer.

PR #235, opened 2026-06-01
fix(muggle-pr-followup): reconcile sweep for slots whose PR went terminal while polling lapsed

stan4git:
## The bug you flagged

The watcher kept "tracking" #234 after it merged. Root cause: **termination is tick-driven.** `contract.md` Step 1 (refresh state) + Step 2 (terminate on `MERGED`/`CLOSED`) *do* poll state every tick — but a slot only finalizes when a tick actually fires and observes the transition. If the tick stream stops first, nothing catches it:

- the recurring `/loop` cron **auto-expires after 7 days**,
- the session ends, or
- the machine is off at merge time.

→ the slot is left un-finalized: no `result.md`, no post-merge cleanup, and a surviving cron would keep polling a dead PR. (#234 merged at 17:27Z, hours after the last tick fired — so no tick was alive to finalize it.)

So it wasn't "doesn't check state" — the poll is correct; the gap is the **lack of a catch-up when polling lapses.**

## Fix

- **`reconcile.md`** — a sweep mode: scan session slots, refresh each tracked PR's live state, and finalize the terminal ones. Finalize-only — it never seeds or re-arms a watcher (re-arming an open PR stays auto-track's job). Idempotent (skips slots that already have `result.md`); per-slot isolated (one `gh` failure never blocks the rest).
- **`finalize.md`** — the termination sequence (mark terminal → `result.md` → log/telemetry → unschedule cron → post-merge cleanup handoff) extracted into one shared file. `contract.md` Step 2 and `reconcile.md` both call it, so there's a single source of truth.
- **Wiring** — reconcile runs at **`auto-track` Step 0** (the natural no-arg re-entry, so a bare `/muggle-pr-followup` now also finalizes orphaned terminal slots), plus an explicit **`reconcile` / `sweep`** route.

## Verification

Docs-only (skill markdown). `verify:plugin` and `verify:contracts` pass; all intra-folder links resolve. No unit tests cover skill `.md`; no E2E (this is the skill definition, not app code).

### DIFF
--- a/plugin/skills/muggle-pr-followup/finalize.md
+++ b/plugin/skills/muggle-pr-followup/finalize.md
@@ -1,0 +1,37 @@
+# Finalize a Terminal PR
+
+The shared termination sequence for a follow-up slot whose PR is `MERGED` or `CLOSED`. Called by [`contract.md`](contract.md) Step 2 (a tick observed the transition) and [`reconcile.md`](reconcile.md) (a sweep found a slot whose polling lapsed before the transition). One slot, run once.
+
+## Inputs
+
+- `<slug>`, `<owner>/<repo>`, `<n>` — the slot's PR.
+- `state` — `MERGED` or `CLOSED`, from a fresh [`../_shared/github-cli-recipes/pr-metadata.md`](../_shared/github-cli-recipes/pr-metadata.md).
+- `mergeCommit` + `mergedAt` when `MERGED`.
+
+## Procedure
+
+### Step 1 — Mark the slot terminal
+
+Rewrite `prs.json[0].state` to `merged` / `closed` ([`state-schemas.md`](state-schemas.md#prsjson)). That state plus the `result.md` written next are the terminal marker — there is no separate flag.
+
+### Step 2 — Write `result.md`
+
+Once, per [`state-schemas.md`](state-schemas.md#resultmd). Pull `cycles_completed`, `pushed_shas`, and `escalated_review_ids` from `last_seen.json`.
+
+### Step 3 — Log and telemetry
+
+Append the terminal line per [`output-templates/watcher-log.md`](output-templates/watcher-log.md). Emit one `tick` event with `terminal: true` per [`../_shared/telemetry-events/pr-followup-tick.md`](../_shared/telemetry-events/pr-followup-tick.md).
+
+### Step 4 — Unschedule the cron
+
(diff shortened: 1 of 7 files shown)


plugin/skills/muggle-pr-followup/finalize.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Step 5 — Post-merge cleanup (`MERGED` only)
+
+Dispatch `/muggle-do` with the cleanup directive as a closing action of the turn — the tick emits exactly one, a reconcile sweep emits one per merged slot:
+
+```
+/muggle-do post-merge cleanup slug=<slug>
+```
+
+`/muggle-do` owns the worktree/branch knowledge and honors the `autoCleanup` gate. Skip on `CLOSED` — unmerged, so leave the branch and any worktree intact.
```
stan4git: wait, why reference muggle-do?


SoC please, and no reverse ref

plugin/skills/muggle-pr-followup/reconcile.md
```diff
@@ -0,0 +1,33 @@
+# Reconcile Procedure
```
stan4git, line 1: Can you please add a rule to this repo, so that all skills' content are one-way dependent, free of reverse dependencies (doc A reference B, B then reference A)

plugin/skills/muggle-pr-followup/finalize.md
stan4git: Addressed in 9871f6a. `finalize.md` is now a pure primitive — mark terminal, write `result.md`, log/telemetry, unschedule the cron, and it dispatches nothing. The `/muggle-do post-merge cleanup` hand-off moved back to `contract.md` Step 2 (the tick orchestrator), where the watcher fires it as a runtime dispatch. So the shared finalize step references no `/muggle-do` and any caller (tick or reconcile) can reuse it.


plugin/skills/muggle-pr-followup/reconcile.md
stan4git: Done in 9871f6a — added `plugin/skills/CLAUDE.md` codifying the rule for all skills: cross-references form a one-way graph; if A references B, no file in B may reference A back. It draws the line you care about — a runtime slash-dispatch (`/muggle-do …`) is an action, not a doc dependency, so a dumb-pipe may still fire its executor; what is forbidden is a procedure file linking to or encoding the other skill. Includes the `muggle-pr-followup` <- `muggle-do` worked example.

I also made this watcher folder comply with the new rule: dropped the reverse `../do/` markdown links from `muggle-pr-followup/CLAUDE.md` and `auto-track.md` (the runtime `/muggle-do` dispatches in `contract.md` stay). Sweeping the remaining `muggle-do` <-> `muggle-pr-followup` doc-link cycle on the `do/` side, plus an enforcing verify-script cycle-detector, is a larger follow-up across all skills — happy to take it as a separate PR. Flagging: a naive enforced check would currently fail on that pre-existing `do/` -> `followup/` <-> `followup/` cycle, so the cleanup needs to land first.

PR #242, opened 2026-06-02
fix(muggle-do): stop double-scheduling the watcher cron after a cycle

stan4git:
## Problem

The watcher's recurring `/loop` cron is created **once** at bootstrap / auto-track / forward-Stage-8 and is the single scheduler for a PR (`contract.md` Step 6 idle path relies on it persisting). But two watcher-dispatched `/muggle-do` modes re-dispatched `/loop` at the end of their cycle:

- `do/address-reviews.md` Step 6 ("Respawn the watcher")
- `do/fix-ci.md` Step 5 ("Update state + respawn")

Each re-dispatch registers a **duplicate** cron on top of the still-live one → the watcher fires twice a minute, and it compounds every cycle.

Observed live: handling a review on `muggle-ai-brain#76`, following Step 6 literally would have stacked a second cron on the bootstrap cron.

## Fix

Both modes now rely on the persistent cron and never re-dispatch `/loop`:

- **PR open:** do nothing — the next tick arrives via the existing cron.
- **PR terminal:** leave finalization to the next watcher tick (`contract.md` Step 2 → `finalize.md`: `result.md` + cron unschedule + post-merge cleanup). Single source of truth for finalize; no duplication.

`fix-ci`'s "CI is the verify loop" behavior is preserved — the next tick re-checks CI on the new SHA, bounded by the per-SHA fix budget.

Markdown-only. Initial cron-creation sites (bootstrap, auto-track, forward Stage 8) are unchanged.

Part of the harness pipeline-integration work — design: `muggle-ai-brain` PR #76.

### DIFF
--- a/plugin/skills/do/address-reviews.md
+++ b/plugin/skills/do/address-reviews.md
@@ -114,7 +114,7 @@ Refresh PR state per [`../_shared/github-cli-recipes/pr-metadata.md`](../_shared
 1. Write `result.md` per [`../muggle-pr-followup/state-schemas.md`](../muggle-pr-followup/state-schemas.md#resultmd).
 2. Do **not** respawn the watcher.
 
-Otherwise, dispatch the next watcher as the last action of this turn:
+Otherwise, dispatch the next watcher as the last action of this turn. The watcher cancelled its own cron when it dispatched this cycle ([`../muggle-pr-followup/contract.md`](../muggle-pr-followup/contract.md) Step 4), so this restart is the single live watcher — never a duplicate:
 
 ```
 /loop 1m /muggle:muggle-pr-followup <slug> <n>
--- a/plugin/skills/do/fix-ci.md
+++ b/plugin/skills/do/fix-ci.md
@@ -40,7 +40,7 @@ Commit per the `fix(ci): <check> — <what>` convention ([`../_shared/pr-followu
 ### Step 5 — Update state + respawn
 
 - Increment `last_seen.ci_fix_attempts[red_sha]`.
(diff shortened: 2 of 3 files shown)


plugin/skills/do/address-reviews.md
```diff
@@ -107,18 +107,12 @@ Invoke [`per-comment-replies.md`](per-comment-replies.md) with the actionable re
 
 Invoke [`resolve-reminder.md`](resolve-reminder.md) once, regardless of whether this round pushed. It scans unresolved threads, finds those whose newest comment is loop-marked (addressed, awaiting resolve), and posts one top-level PR comment nudging the reviewer to resolve them — or stays silent if there are none. Threads with a newer human comment were already pulled into this round's work set in Step 1(b).
 
-### Step 6 — Respawn the watcher
+### Step 6 — Hand back to the watcher
```
stan4git: I don't like this at all.

Running Watcher should be stopped once dev cycler kicks in to address comments/ci/lint problem. It should be a single thread process.


Once dev cycle done (test etc), start a new watcher cycle

stan4git: Reworked to your model — the watcher is now **single-threaded**.

- On dispatch (`contract.md` Step 4 for reviews, Step 5 for CI), the watcher **cancels its own cron** before handing off to `/muggle-do`. No tick fires while a cycle runs.
- `/muggle-do` **respawns** the watcher when the cycle finishes (`address-reviews` Step 6, `fix-ci` Step 5). Exactly one cron per PR, ever.

This replaces my earlier "persistent cron, never re-dispatch" approach (which you correctly rejected — it let a tick overlap a running cycle). (e08de30)

PR #263, opened 2026-06-09
refactor(muggle-test): router + per-path execution files

stan4git:
## What

Make `muggle-test` a thin **router**. The local and remote execution paths each get their own process file; the router holds only mode-agnostic steps.

- `muggle-test/SKILL.md` (router, 417→317 lines): selection (Steps 1–6f) → **Step 7** (fetch test-case details *once*, then dispatch) → 7C failures → 8–10. All post-run steps operate on a **uniform runs list** `[{ testCaseId, mode, runId|runtimeId, status, viewUrl? }]`.
- `muggle-test/execute-local.md` — local process: prepare → URL/visibility pre-flight → dev loop → collect → publish.
- `muggle-test/execute-remote.md` — remote process: ask URL → trigger workflows → monitor.

## Why

The old 7A/7B duplicated "Fetch test case details" verbatim and split Step 8 by mode. Now shared steps live once in the router (or shared `_shared/` files), and each path defines its own process.

## De-duplication

- Fetch-test-case-details: deduped — runs once in the router before dispatch.
- Step 8 open-results: unified (keyed on whether runs carry published viewUrls).
- Failures (`debug-failed-run.md`) and local mechanics (`_shared/dev-loop/*`) were already shared; the path files reference them, one-way (path files never link back to the router).

Per-path **report tables stay per-path** — different columns is divergence, not duplication.

### DIFF
--- a/plugin/skills/muggle-test/SKILL.md
+++ b/plugin/skills/muggle-test/SKILL.md
@@ -42,11 +42,11 @@ Gates run per `preference-gates/README.md`.
 |------------|------|-------------------|
 | `autoLogin` | 3 | Reuse saved credentials when auth is required |
 | `autoSelectProject` | 4 | Reuse last-used Muggle Test project for this repo |
-| `autoSelectLocalHost` | 7A | Reuse last-used local dev server URL for this repo |
+| `autoSelectLocalHost` | execute-local | Reuse last-used local dev server URL for this repo |
 | `autoDetectChanges` | 2 | Scan local git changes and map to affected test cases |
 | `defaultExecutionMode` | 1 | Default to local or remote test execution |
-| `autoPublishLocalResults` | 7A | Upload local results to Muggle Test cloud after run |
-| `showElectronBrowser` | 7A | Show the Electron browser window during local test execution (vs. run headless) |
+| `autoPublishLocalResults` | execute-local | Upload local results to Muggle Test cloud after run |
+| `showElectronBrowser` | execute-local | Show the Electron browser window during local test execution (vs. run headless) |
 | `postPRVisualWalkthrough` | 9 | Post visual walkthrough to PR after results are available |
 | `autoCreatePR` | 9 (if no PR) | Auto-create the PR when posting the walkthrough has no PR to target |
 | `autoWatchPR` | 9.5 (if a PR exists) | Start a `muggle-pr-followup` watcher on the PR after the run |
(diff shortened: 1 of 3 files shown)


plugin/skills/muggle-test/SKILL.md
```diff
@@ -42,11 +42,11 @@ Gates run per `preference-gates/README.md`.
 |------------|------|-------------------|
 | `autoLogin` | 3 | Reuse saved credentials when auth is required |
 | `autoSelectProject` | 4 | Reuse last-used Muggle Test project for this repo |
-| `autoSelectLocalHost` | 7A | Reuse last-used local dev server URL for this repo |
+| `autoSelectLocalHost` | execute-local | Reuse last-used local dev server URL for this repo |
 | `autoDetectChanges` | 2 | Scan local git changes and map to affected test cases |
 | `defaultExecutionMode` | 1 | Default to local or remote test execution |
-| `autoPublishLocalResults` | 7A | Upload local results to Muggle Test cloud after run |
-| `showElectronBrowser` | 7A | Show the Electron browser window during local test execution (vs. run headless) |
+| `autoPublishLocalResults` | execute-local | Upload local results to Muggle Test cloud after run |
```
stan4git, line 48: I thought this is not needed??

plugin/skills/muggle-test/execute-local.md
```diff
@@ -0,0 +1,71 @@
...
+Caller glue:
+- `mode` per test case is the input from the router; `localUrl` from the pre-flight question; `showUi` from the `showElectronBrowser` resolution.
+- `cwd` is the input from the router — it drives the cross-worktree single-flight lock so concurrent muggle-test runs from different branches serialize.
+- On a failed run, continue the batch; the router routes failures through the debug path after this path returns.
+
+## Collect results
+
+Fetch every `runId` per [`../_shared/dev-loop/failures.md`](../_shared/dev-loop/failures.md), reading structured fields and [interpreting failures](../_shared/dev-loop/failures.md) — never `execute`'s stdout tail. Issue the `muggle-local-run-result-get` calls in parallel; use `Error` as the headline for failures.
+
+## Publish each run to cloud (gated by `autoPublishLocalResults`)
```
stan4git, line 44: I feel this should never needed (electron app does it already?)

plugin/skills/muggle-test/SKILL.md
stan4git: Verified it's needed. The Electron app saves results locally (Firestore + Firebase screenshots) — that's what the debug path reads — but it does NOT publish to the cloud dashboard. `muggle-local-publish-test-script` is the only path that uploads the full action script + cloud-hosted screenshots and returns the dashboard `viewUrl` + cloud `actionScriptId` (which muggle-feedback anchors on). For replay runs the Electron side skips cloud test-script creation entirely (export.ts:264 is explore-mode only). So the publish step stays.

plugin/skills/muggle-test/execute-local.md
stan4git: Traced it — the Electron app does not publish local runs to the cloud. It writes the action script to local Firestore and stores screenshots in Firebase, but the cloud dashboard record + `viewUrl` come only from the works-side publish (`POST /muggle-test/local-run/upload`), and for replay runs the Electron side skips cloud test-script creation altogether. Same split as the screenshots: electron = local; cloud dashboard / PR walkthrough / feedback-anchor = this step. Keeping it.

stan4git: Correction — you're right, and I had the layer wrong. Publishing is needed, but it belongs in the **studio → backend** path, not this skill step. Plan: the studio (electron app) calls `POST /muggle-test/local-run/upload` itself on run completion (regen + replay) and the IDs flow back via the run result; then this publish step + the `autoPublishLocalResults` gate + the `muggle-local-publish-test-script` tool are removed. Sequenced studio-first so local runs aren't orphaned. Tracked in design `studio-publishes-local-runs`. This step stays here only until the studio change lands.

plugin/skills/muggle-test/SKILL.md
stan4git, line 48: The result is going to be published from studio, not via mcp

stan4git: Correct — and that's exactly what the stacked #267 (`refactor/works-drops-skill-publish`) does: it removes the `autoPublishLocalResults` gate, the publish step, and the `muggle-local-publish-test-script` tool, repointing the skill to the cloud identifiers the **studio** stamps on the run result (`viewUrl` / `cloudTestScriptId` / `cloudActionScriptId`).

#263 keeps this row only as the Phase-1 fallback, per the `studio-publishes-local-runs` design's hard sequencing constraint — works must not drop its own publish until the studio that publishes is deployed (merge order: #277 deploy → #263 → #267). Happy to fold the removal into #263 instead if you'd rather collapse the two-PR split.

PR #268, opened 2026-06-09
chore(skills): broaden muggle-do auto-trigger description

stan4git:
## What

Replaces muggle-do's thin frontmatter `description` (247 chars) with an enriched, intent-focused one (968 chars) that better auto-triggers on natural-language "build/implement/fix X → open a PR" requests, and explicitly scopes it away from its siblings.

Only `plugin/skills/muggle-do/SKILL.md` changes — one line.

## Why (eval-driven)

Ran the skill-creator trigger harness against **all three** workflow skills (muggle-do / muggle-test / muggle-pr-followup). Real muggle skills were isolated via a separate `CLAUDE_CONFIG_DIR` so the installed copies couldn't steal the trigger match. Eval sets: 16–19 queries each, ~half **hard cross-skill negatives** (the real risk is these three colliding).

Baseline (held-out test split): all three **under-trigger** (recall 11–33%) but **never over-trigger** (precision 100%). The bottleneck is mostly dispositional — for muggle-do, its trigger phrases *are* generic coding requests the model just executes directly.

| skill | decision | rationale |
|---|---|---|
| **muggle-do** | **enriched (this PR)** | thinnest desc; across 3 runs enriched beat original on raw recall every time (~2–5×, 4%→9%), **0 new false-fires** |
| muggle-test | unchanged | all 5 auto-rewrites *reduced* raw recall (4→3); already best |
| muggle-pr-followup | unchanged | flat (6/18 either way); already keyword-saturated |

The gain is real but **marginal** — wording is a weak lever here; absolute auto-trigger stays modest because the model tends to just do coding tasks. This change is low-risk hygiene: free recall improvement + clearer disambiguation + consistency with sibling descriptions.

## Contracts preserved

- literal **`muggle do`** keyword
- internal **`address-reviews`** directive (dispatched by the muggle-pr-followup watcher)
- one-way reference rule respected (muggle-do references muggle-test / muggle-pr-followup *downward*)

Goes live after a `@muggleai/works` plugin release.


plugin/skills/muggle-do/SKILL.md
```diff
@@ -1,6 +1,6 @@
...
+description: Unified Muggle AI workflow entry point. Use to take a feature request, spec, design, bug, or ticket and carry it all the way to a pull request autonomously — write the code, test it, and open the PR with little further input. Trigger whenever the user wants something built, implemented, added, or fixed and a PR opened, raised, or put up at the end, e.g. "implement rate limiting and raise a PR", "build the CSV export button and put up a PR", "take this Figma/Jira ticket and run with it, ping me when the PR is up", "fix this bug and open a pr", or simply "muggle do …". The signal is a build/implement/fix request paired with opening a PR, handed off for hands-off execution. Also handles the internal `address-reviews` directive dispatched by the muggle-pr-followup watcher when new reviews land on a PR. Not for testing or validating code the user already wrote (use muggle-test), and not for ongoing watching of a PR's incoming reviews (use muggle-pr-followup).
```
stan4git: "Unified Muggle AI workflow entry point." this is not helpful at all. Kepe it tight

stan4git: Done — dropped the "Unified Muggle AI workflow entry point" opener and tightened the whole description (714 chars, was 968). It now leads with the action ("Take a feature request… carry it autonomously to an opened PR"). Pushed in 3ba01c4.

PR #279, opened 2026-06-13
feat(mcp): lane-aware test-script-list (local-lane fetch for replay)

stan4git:
## What
1. Adds an optional `runEnvironmentType` to **`muggle-remote-test-script-list`**, mapped to the backend's `type` query filter (prompt-service #624). A local replay can now fetch the **local-lane** (`type=local`, localhost) test script instead of the remote canonical one.
2. Wires that lane param into the **skills** at every place they resolve which script to replay, so a local run reads the localhost-lane script and a remote run the deployed-lane script.

## Why (supersedes the original approach)
This PR originally added a `productionUrl` **side-by-side** with `url` in the action-script payload, feeding a studio-side runtime **origin-pin** (teaching-service #287). That was the wrong layer — the lane is fixed by runSettings and **not interchangeable**, so storing both lane URLs on one script and rewriting origins at run time is confusing and fragile. Per review, both of those changes were reverted (teaching #287 closed), and the per-lane URL now lives on **separate per-lane test scripts**:

- **prompt-service #624**: the local-run upload persists the local-lane script with the **localhost** URL via a lane-keyed get-or-create (no clobber of the remote canonical), and exposes a `type` filter on the list endpoint.
- **this PR**: the MCP list tool forwards the lane, and the skills pass it on every replay-script lookup.

## Skill wiring (read-side lane scoping)
- `_shared/failure-mode-handling.md` §A (shared replay-vs-regen classification): scope the list call to the run's lane — `local` in Mode A, `remote` in Mode B.
- `muggle-test` 6f + `execute-remote.md`, `do/e2e-acceptance.md` (local E2E), `muggle-test-feature-local`, `muggle-browser-task`: pass `runEnvironmentType` at each resolution point.
- `_shared/dev-loop/run.md`: notes the replay script is the lane-scoped one from §A.

## Result
With per-lane scripts, the existing `rewriteActionScriptUrls` becomes a benign localhost→localhost port-adaptation rather than a lane-replacing public→localhost rewrite — the agent's context is always the lane's own URL.

## Tests
`tsc --noEmit` clean. Schema change is an optional field; skill edits are markdown. CI runs the suite.

## Deploy note
Skill edits live in `plugin/skills/` — they reach users only after a plugin build/release (`@muggleai/works` npm publish). The MCP param change ships the same way.

### DIFF
--- a/packages/mcps/src/mcp/tools/e2e/tool-registry.ts
+++ b/packages/mcps/src/mcp/tools/e2e/tool-registry.ts
@@ -452,7 +452,7 @@ const testScriptTools: IQaToolDefinition[] = [
   {
     name: "muggle-remote-test-script-list",
     description:
-      "List test scripts for a project, optionally filtered by test case. Returns up to 10 items per page by default (max 100). Response includes pagination metadata (totalCount, totalPages, hasMore) — check `hasMore` to decide whether to fetch additional pages.",
+      "List test scripts for a project, optionally filtered by test case and by environment lane. Returns up to 10 items per page by default (max 100). Response includes pagination metadata (totalCount, totalPages, hasMore) — check `hasMore` to decide whether to fetch additional pages. Pass runEnvironmentType: 'local' to fetch the local-lane (localhost) script for a local replay, or 'remote' for the deployed script.",
     inputSchema: schemas.TestScriptListInputSchema,
     mapToUpstream: (input) => {
       const data = input as z.infer<typeof schemas.TestScriptListInputSchema>;
@@ -462,6 +462,7 @@ const testScriptTools: IQaToolDefinition[] = [
         queryParams: {
           projectId: data.projectId,
           testCaseId: data.testCaseId,
+          type: data.runEnvironmentType,
           page: data.page,
           pageSize: data.pageSize,
           sortBy: data.sortBy,
--- a/packages/mcps/src/mcp/tools/e2e/tool-registry.ts
+++ b/packages/mcps/src/mcp/tools/e2e/tool-registry.ts
@@ -462,7 +462,7 @@ const testScriptTools: IQaToolDefinition[] = [
         queryParams: {
           projectId: data.projectId,
           testCaseId: data.testCaseId,
-          type: data.runEnvironmentType,
+          runEnvironmentType: data.runEnvironmentType,
           page: data.page,
           pageSize: data.pageSize,
           sortBy: data.sortBy,
--- a/packages/mcps/src/mcp/tools/e2e/tool-registry.ts
+++ b/packages/mcps/src/mcp/tools/e2e/tool-registry.ts
@@ -452,7 +452,7 @@ const testScriptTools: IQaToolDefinition[] = [
   {
     name: "muggle-remote-test-script-list",
     description:
-      "List test scripts for a project, optionally filtered by test case and by environment lane. Returns up to 10 items per page by default (max 100). Response includes pagination metadata (totalCount, totalPages, hasMore) — check `hasMore` to decide whether to fetch additional pages. Pass runEnvironmentType: 'local' to fetch the local-lane (localhost) script for a local replay, or 'remote' for the deployed script.",
+      "List test scripts for a project, optionally filtered by test case and by environment lane. Returns up to 10 items per page by default (max 100). Response includes pagination metadata (totalCount, totalPages, hasMore) — check `hasMore` to decide whether to fetch additional pages.",
     inputSchema: schemas.TestScriptListInputSchema,
     mapToUpstream: (input) => {
       const data = input as z.infer<typeof schemas.TestScriptListInputSchema>;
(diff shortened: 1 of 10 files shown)


packages/mcps/src/mcp/local/services/action-script-builders.ts
```diff
@@ -62,6 +62,13 @@ export function buildGenerationActionScript(params: {
       actionScriptId: params.localTestScriptId,
       workflowRunId: params.runId,
       url: params.localUrl,
+      // Lane endpoints handed to the studio so it can pin every in-run
+      // navigation to the active lane's origin: `url` is the local target the
+      // run executes against, `productionUrl` is the cloud origin the test case
+      // was authored for. On a local run the agent can still emit a navigation
+      // to the production origin (it appears in the goal/screenshots); the
+      // studio rewrites any such production-origin hop back to `url`'s origin.
+      productionUrl: params.testCase.url,
```
stan4git: lets not put env dependencies into data model

stan4git: Agreed — removed it. I reverted the `productionUrl` side-by-side here, so no env/lane dependency is crammed onto one script's data model. The lane now lives on **separate per-lane test scripts**: prompt-service #624 makes the local-run upload persist the local-lane script with the localhost URL (and keeps remote on its own script) via a lane-keyed get-or-create — no clobber, no schema change. This PR will be repurposed for the works-side wiring: lane-aware `test-script-get`/`list` so a local replay fetches the `type=local` script, dropping the brittle URL host-swap rewrite.

plugin/skills/muggle-browser-task/SKILL.md
```diff
@@ -74,6 +74,7 @@ Tell the user: `Created test case: <id>`.
 Load and call `muggle-remote-test-script-list` with:
 - `projectId`: from Step 2
 - `testCaseId`: from Step 4
+- `runEnvironmentType`: `"local"` when `localUrl` (Step 1) is a loopback host (`localhost`, `127.0.0.1`, `[::1]`), else `"remote"` — replay runs locally against `localUrl`, so this resolves the script on the lane that URL belongs to.
```
stan4git: i dont like using "localUrl", it should br exactly likr remote only diff is env type local, it should has url prop just like remote

stan4git: Addressed in a19b1ef: dropped the loopback-host inspection. muggle-browser-task executes in the local browser, so it now passes `runEnvironmentType: "local"` directly — symmetric with the remote flow (which passes `"remote"`). The lane is the explicit env type; the target URL is just the lane's url and no longer decides the lane.


stan4git:
**Review feedback addressed in `a19b1ef`** — 1 thread resolved on my end, awaiting your resolve:

- `muggle-browser-task/SKILL.md` — lane is now explicit `runEnvironmentType: "local"` (symmetric with the remote flow); dropped the `localUrl` loopback-host inference.

CI green (10/10) on the prior head; this is a docs-only skill change. Please resolve the thread if it looks right.


packages/mcps/src/mcp/tools/e2e/tool-registry.ts
```diff
@@ -452,7 +452,7 @@ const testScriptTools: IQaToolDefinition[] = [
   {
     name: "muggle-remote-test-script-list",
     description:
-      "List test scripts for a project, optionally filtered by test case. Returns up to 10 items per page by default (max 100). Response includes pagination metadata (totalCount, totalPages, hasMore) — check `hasMore` to decide whether to fetch additional pages.",
+      "List test scripts for a project, optionally filtered by test case and by environment lane. Returns up to 10 items per page by default (max 100). Response includes pagination metadata (totalCount, totalPages, hasMore) — check `hasMore` to decide whether to fetch additional pages. Pass runEnvironmentType: 'local' to fetch the local-lane (localhost) script for a local replay, or 'remote' for the deployed script.",
```
stan4git: why you has to call out a few param in general description? i odnt like making param sepcial

stan4git: Addressed in 556780c: dropped the `runEnvironmentType` value call-out ('local' / 'remote') from the general description. The lane is now mentioned only as a filter dimension alongside test case ("optionally filtered by test case and by environment lane"), and the value semantics stay on `RunEnvironmentInputSchema.describe()` — where every other param documents itself. No param singled out in the general description.

If you'd also like "and by environment lane" dropped entirely (treating it differently from the existing "by test case" mention), say so and I'll remove it too.


stan4git:
This thread is addressed and still open — mark it resolved if satisfied, or reply if more is needed:
- [`packages/mcps/src/mcp/tools/e2e/tool-registry.ts` — "why call out a few param in general description?"](https://github.com/multiplex-ai/muggle-ai-works/pull/279#discussion_r3410506278)

PR #280, opened 2026-06-15
feat(guardrails): hard-enforce E2E run + CLI-rendered PR reports

stan4git:
## Problem

The E2E acceptance gate and the PR walkthrough were **advisory only**. The model could (and did) skip the E2E run and hand-write freeform "test reports" that bypass the deterministic `muggle build-pr-section` renderer. The Stop gate even set `e2eRun=true` the moment it *reminded* — so it fired once, lied that E2E had run, and never re-armed.

## What changed

**E2E run — now enforced (`Stop` → `decision: block`)**
- The gate refuses to end the turn until a real E2E run is recorded, instead of an ignorable `additionalContext` nudge.
- Removed the self-marking lie: `e2eRun` is set **only** when an actual run is detected (`guardrail-record-tests`).
- Escape hatch: releases to advisory after `MAX_E2E_BLOCKS` (3) blocks so a genuinely un-runnable E2E (no app/services/PR) can't trap the turn.
- E2E runs driven by the **muggle execute/replay MCP tools** are now recorded too (new `PostToolUse` matcher), not just Bash `muggle …` — otherwise the block could never be satisfied.

**Report formatting — now enforced (`PreToolUse` → `permissionDecision: deny`)**
- `muggle build-pr-section` prepends an invisible sentinel (`<!-- muggle-pr-section:v1 -->`) to `body` and `comment`.
- New `guardrail-report-format` denies a `gh pr comment|create|edit` whose body **reads like an E2E results report** (acceptance-results heading / pass-fail tally / per-test status list + Muggle/E2E context) but **lacks the sentinel**.
- Fails **open** on anything it can't positively read as a hand-written report: the sanctioned `jq -r '.body' … | gh … --body-file -` path (sentinel surfaced via the artifact), ordinary comments, and feature PR descriptions that merely mention E2E all pass.

Cursor has no block/deny primitive, so both gates degrade to advisories there.

## Tunable knob for review
`looksLikeE2EReport` (src/guardrails/reportGate.ts) is the precision/recall dial — it keys on report *structure*, not just the words "e2e"/"muggle", to avoid flagging normal PR descriptions. Worth a look if you want it tighter or looser.

## Verification
- `vitest run` — **382 passing** (25 files), incl. new `reportGate` + extended `emit`/`shouldRunE2E` suites.
- `tsc --noEmit` clean; eslint clean on changed files.
- Bundled `guardrails.mjs` smoke-tested end-to-end: report-gate denies a hand-written report / allows sentinel + ordinary comments; E2E gate blocks 3× then releases, and a recorded MCP E2E run satisfies it.

Docs: `plugin/hooks/README.md` updated with an **advise vs enforce** section and a strength column.

### DIFF
--- a/plugin/scripts/guardrails.mjs
+++ b/plugin/scripts/guardrails.mjs
@@ -66,9 +66,9 @@ function shouldRunE2E(state) {
 }
 function e2eGateDecision(state, maxBlocks = MAX_E2E_BLOCKS) {
   const blockCount = state.e2eBlockCount ?? 0;
-  if (!shouldRunE2E(state)) return { action: "none", blockCount };
-  if (blockCount >= maxBlocks) return { action: "release", blockCount };
-  return { action: "block", blockCount: blockCount + 1 };
+  if (!shouldRunE2E(state)) return { action: "none" /* None */, blockCount };
+  if (blockCount >= maxBlocks) return { action: "release" /* Release */, blockCount };
+  return { action: "block" /* Block */, blockCount: blockCount + 1 };
 }
 
 // src/guardrails/detectBuildIntent.ts
@@ -200,7 +200,7 @@ function recordTests() {
 function e2eGate() {
   const state = readState(sessionId);
   const decision = e2eGateDecision(state);
-  if (decision.action === "none" || decision.action === "release") return "{}";
+  if (decision.action === "none" /* None */ || decision.action === "release" /* Release */) return "{}";
   state.e2eBlockCount = decision.blockCount;
   writeState(state);
   const reason = `Do not end the turn yet. Unit tests passed this session but no E2E acceptance run has happened. Per the autoE2ETest preference (default: always), run change-driven E2E now via /muggle:muggle-test, then finish. If E2E genuinely cannot run here (no app, services down, no PR), say so explicitly to the user \u2014 this gate releases after ${MAX_E2E_BLOCKS} attempts.`;
--- a/plugin/scripts/guardrails.mjs
+++ b/plugin/scripts/guardrails.mjs
@@ -64,6 +64,16 @@ var MAX_E2E_BLOCKS = 3;
 function shouldRunE2E(state) {
   return state.unitTestsGreen === true && state.e2eRun !== true;
 }
+function applyRecordedRun(state, run) {
+  let next = state;
+  if (run.unitTestPassed) {
+    next = { ...next, unitTestsGreen: true, e2eRun: false, e2eBlockCount: 0 };
+  }
+  if (run.e2eRan) {
+    next = { ...next, e2eRun: true };
+  }
+  return next;
+}
 function e2eGateDecision(state, maxBlocks = MAX_E2E_BLOCKS) {
   const blockCount = state.e2eBlockCount ?? 0;
   if (!shouldRunE2E(state)) return { action: "none" /* None */, blockCount };
@@ -185,16 +195,11 @@ Per the autoWatchPR preference, a muggle-pr-followup watcher should handle its i
 function recordTests() {
   const cmd = input.tool_input?.command ?? "";
   const state = readState(sessionId);
-  let changed = false;
-  if (isTestCommand(cmd) && testsPassed(input)) {
-    state.unitTestsGreen = true;
-    changed = true;
-  }
-  if (isE2ERun(input)) {
-    state.e2eRun = true;
-    changed = true;
-  }
-  if (changed) writeState(state);
+  const next = applyRecordedRun(state, {
+    unitTestPassed: isTestCommand(cmd) && testsPassed(input),
+    e2eRan: isE2ERun(input)
+  });
+  if (next !== state) writeState(next);
   return "{}";
 }
 function e2eGate() {
--- a/plugin/scripts/guardrails.mjs
+++ b/plugin/scripts/guardrails.mjs
@@ -1,5 +1,5 @@
 import { readFileSync, existsSync, mkdirSync, writeFileSync } from 'fs';
-import { join } from 'path';
+import { isAbsolute, resolve, join } from 'path';
 import { homedir } from 'os';
 
 // src/guardrails/cli.ts
@@ -60,8 +60,15 @@ function isE2ERun(input2) {
 }
 
 // src/guardrails/shouldRunE2E.ts
+var MAX_E2E_BLOCKS = 3;
 function shouldRunE2E(state) {
   return state.unitTestsGreen === true && state.e2eRun !== true;
+}
+function e2eGateDecision(state, maxBlocks = MAX_E2E_BLOCKS) {
+  const blockCount = state.e2eBlockCount ?? 0;
+  if (!shouldRunE2E(state)) return { action: "none", blockCount };
+  if (blockCount >= maxBlocks) return { action: "release", blockCount };
+  return { action: "block", blockCount: blockCount + 1 };
 }
 
 // src/guardrails/detectBuildIntent.ts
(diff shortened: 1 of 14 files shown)


plugin/scripts/guardrails.mjs
```diff
@@ -60,9 +60,16 @@ function isE2ERun(input2) {
 }
 
 // src/guardrails/shouldRunE2E.ts
+var MAX_E2E_BLOCKS = 3;
 function shouldRunE2E(state) {
   return state.unitTestsGreen === true && state.e2eRun !== true;
```
stan4git, line 65: after addressing comment, new e2e is rrquired

src/guardrails/shouldRunE2E.ts
```diff
@@ -1,5 +1,30 @@
 import type { GuardrailState } from "./types.js";
 
+export const MAX_E2E_BLOCKS = 3;
+
 export function shouldRunE2E(state: GuardrailState): boolean {
   return state.unitTestsGreen === true && state.e2eRun !== true;
 }
+
+export type E2eGateAction = "block" | "release" | "none";
```
stan4git: enum

src/guardrails/shouldRunE2E.ts
```diff
@@ -1,5 +1,30 @@
 import type { GuardrailState } from "./types.js";
 
+export const MAX_E2E_BLOCKS = 3;
+
 export function shouldRunE2E(state: GuardrailState): boolean {
   return state.unitTestsGreen === true && state.e2eRun !== true;
```
stan4git, line 6: same as other comment

src/guardrails/shouldRunE2E.ts
stan4git: Addressed in 5f2b080: converted `E2eGateAction` from a `"block" | "release" | "none"` union literal to a TS string enum (`E2eGateAction.Block/Release/None`); updated `e2eGateDecision` returns, the Stop-gate comparison in cli.ts, and the tests.


src/guardrails/shouldRunE2E.ts
stan4git: Addressed in 5f2b080: same as the other thread — `E2eGateAction` is now a string enum and every call site references the enum members instead of raw string literals.


plugin/scripts/guardrails.mjs
stan4git: Addressed in 5f2b080: change is unit-verified (382 tests pass, typecheck + lint clean). On E2E — this PR records **SKIPPED**, not a run: `muggle-ai-works` is a CLI/MCP plugin with no web dev server, so there is no browser surface to acceptance-test. The change-driven E2E gate treats "no testable surface" as a clean SKIP. (The E2E *enforcement* this PR adds fires on repos that DO have an app/dev-server.)


stan4git:
These threads are addressed in `5f2b080` and still open — mark them resolved if satisfied, or reply if more is needed:
- `shouldRunE2E.ts` — "enum" → `E2eGateAction` is now a string enum
- `shouldRunE2E.ts` — "same as other comment" → same enum conversion, all call sites updated
- `guardrails.mjs` — "new e2e is required" → E2E recorded **SKIPPED** (no web surface in this CLI/MCP repo); unit suite + typecheck + lint green


plugin/scripts/guardrails.mjs
stan4git: I meant, what e2erun be after watcher ran muggle-do

stan4git: When the watcher dispatched `/muggle-do` for the earlier review (the enum change), the E2E acceptance stage recorded **SKIPPED** — no browser E2E ran.

Why: this watcher session is **poll-only** (no E2E validation context was seeded at bootstrap), because `muggle-ai-works` is a CLI/MCP plugin repo with no web dev server / browser-testable surface. A poll-only session yields a clean `SKIPPED` verdict by design (`do/e2e-acceptance.md` Step 0 + bootstrap Step 6.5) — that is **not** a failure. The change was verified instead by `tsc --noEmit` + the 382-test unit suite + eslint.

The E2E *enforcement* this PR adds (the Stop-gate that blocks until a real `muggle-test` run is recorded) fires on repos that actually have an app/dev-server to acceptance-test; here there is nothing to drive a browser against.


stan4git: I meant, this change is to harden e2e engagne logic after poll, i was referring to the state after user finish a mdo cycle

stan4git: Got it — and apologies the earlier replies talked past your point.

Restating correctly: this PR's hardening targets the **post-poll / post-mdo-cycle state** — the moment after a `/muggle-do` cycle finishes with unit tests green but no E2E run yet. That's exactly what the new `Stop`→`block` E2E gate covers: it refuses to end the turn until a real E2E run is recorded (`shouldRunE2E` fires on `unitTestsGreen && !e2eRun`), and the new `guardrail-record-tests` PostToolUse now also records MCP `muggle execute/replay` runs — so a watcher-driven cycle can actually satisfy the gate instead of being stuck.

The earlier SKIPPED tangent was about *this* CLI repo having no browser surface, which wasn't your question.

Does that match your intent, or do you want the gate hardened further for that state?


stan4git: so what haapens if a round of muggle-do with e2e ends with watcher, then gets a new round of comments

stan4git: Short answer: in a repo where E2E **actually runs**, the new round's E2E is **not** re-enforced — that's a real gap. (In *this* CLI repo it doesn't bite: E2E is always SKIPPED, so `e2eRun` is never set and the gate just resolves harmlessly each round. But your question is the "round *with* e2e" case — i.e. an app repo where E2E does run.)

Why: `e2eRun` and `unitTestsGreen` in the guardrail state (`~/.muggle-ai/guardrails/<session_id>.json`) are only ever set `true` in `recordTests` — never reset. The whole `/loop` watcher runs under **one** Claude `session_id`, so once round‑1's E2E sets `e2eRun=true`, it stays `true` for the life of the session. Round‑2's `Stop` gate then:

```
shouldRunE2E = unitTestsGreen && e2eRun !== true   →   true && !true   →   false
e2eGateDecision → "none"   →   Stop gate does NOT block
```

So the second round can finish unit‑green with **no fresh E2E**. And it's broader than just cross‑round: once *any* real E2E runs in the session, the latch stays satisfied for every later round — sticky for the whole watcher lifetime.

Fix — re‑arm the latch at the code‑change boundary. Simplest in‑scope option: in `recordTests`, when a fresh unit‑test pass is recorded, reset `e2eRun=false` (and clear `e2eBlockCount`), so the latest green unit run re‑requires an E2E run after it. Order within a round still holds: code → unit re‑arms → gate blocks → E2E sets `e2eRun=true` → gate clears; round‑2 re‑arms again. I'd add a `shouldRunE2E.test.ts` case asserting the cross‑round reset.

Want me to apply that, or anchor the reset to a different boundary (reset on PR push, or at the start of each address‑reviews cycle)?


stan4git: yes. apply it

stan4git: Addressed in f59dd20: re-armed the E2E gate latch. `recordTests` now folds each run through the new `applyRecordedRun` reducer, which resets `e2eRun=false` and `e2eBlockCount=0` on every fresh unit-test pass — so a new muggle-do round (code → unit-green) re-requires E2E instead of riding the session's first E2E forever. Added a `shouldRunE2E.test.ts` cross-round case asserting the reset (full suite 386 green; typecheck + eslint clean). The bundled `plugin/scripts/guardrails.mjs` was regenerated in the same commit.

PR #315, opened 2026-07-08
fix(pr-followup): 5m poll when blocked; harden cron lifecycle + guarantee respawn

stan4git:
Draft for review. Three design gaps in the `muggle-pr-followup` dumb-pipe watcher, fixed additively — no rearchitecting, no reverse dependency on `/muggle-do`.

## Concern A — a PR blocked pending human input

When a PR can't progress without the user, the 1-minute cron fired forever with zero visible progress (a real PR accrued 1000+ idle ticks over days — blocked on a semantic rebase conflict already in `conflict_escalated_shas` plus an external staging deploy).

Per review, the watcher now **backs off to a `5m` poll while blocked and reminds** — it slows but never stops, and restores the `1m` cadence the instant the block clears.

- **Flag + slow + remind** (`contract.md` Step 7): when an idle tick is a *durable human-block* — `head_sha ∈ conflict_escalated_shas` (`conflict_escalated`), `head_sha ∈ ci_escalated_shas` (`ci_escalated`), or an ambiguous review in `escalated_review_ids` with an empty actionable set (`reviews_escalated`) — it records a `blocked` object in `last_seen.json`, swaps its cron `1m→5m`, and emits a **one-line owner reminder**: the pending act plus a reference back to the decision context. Template: `output-templates/blocked-reminder.md`.
- **Remind or resume** (`contract.md` Step 2.5): every blocked tick recomputes a fingerprint `{ head_sha, latest_review_id, ci_digest }`, re-emits the reminder, and — the instant any component moves (a new push, review, or CI/deploy change) — restores the `1m` cron and falls straight through to the normal evaluation, same tick.
- **Cadence swap** (`blocked-tick.md`): a cancel-then-create (`cancel-cron.md` + `CronCreate`), so the slot is never left cron-less. `cron.json.interval` and the `tick` event carry `"1m" | "5m"` to track the live cadence.

New: `blocked` in `last_seen.json`, `blocked`/`reminded`/`interval` tick-telemetry fields, a `blocked reason=…` `followup.log` line, and the `blocked-reminder.md` template.

## Concern B — orphaned watcher crons

Watcher crons survive session continue / compaction and go **invisible to `CronList`**, so teardown that can only find crons via `CronList` can never kill the orphan — it re-fires until the 7-day `/loop` expiry.

- **Durable id** (`record-cron-id.md`): each tick self-records its `/loop` cron id to `cron.json` *while `CronList` can still see it*, so the id stays a valid `CronDelete` target after the blindness sets in. Bootstrap seeds `cron.json`.
- **Delete by id first** (`cancel-cron.md`): teardown `CronDelete`s the recorded id first (survives blindness), then falls back to the `CronList` two-arg match.
- **Reconcile orphan sweep** (`reconcile.md` Step 3.5): deletes any `pr-followup` cron whose session slot is gone or whose PR is terminal.

## Concern C — a silently-stopped poller

The watcher cancels its own cron the moment it dispatches a `/muggle-do` cycle, relying on the cycle to respawn it at the end. Two escalation paths dropped that respawn — `address-reviews` Step 0 (rebase escalation) and `fix-ci` Step 6 (fix budget spent) — leaving the slot with no cron and no next tick. The poller stopped silently and stayed stopped. (This is the "why did the poller stop after a wave of comments?" report.)

- **Shared respawn helper** (`do/respawn-watcher.md`, new): the single, guaranteed restart used by every watcher-dispatched mode. Respawn is the last action on *every* open-PR exit — happy path and every escalation — with a terminal PR (merged/closed) the only skip. `address-reviews`, `fix-ci`, and `resolve-conflicts` all route through it; the two dropped-respawn holes now respawn.
- **Reconcile recovery net** (`reconcile.md` Step 3.6): re-arms any open slot whose watcher went silent (no tick within a 15-minute staleness window). Guards against a *future* dropped respawn from any cause (a crash mid-cycle), not just the two known holes. Staleness-gated so a live cron — even one `CronList` has gone blind to — is never doubled.

## Tests

Drift-lint `src/test/skills/pr-followup-blocked-cron-lint.test.ts` verifies the cross-file wiring for all three features — the `5m` blocked backoff and cron swap, the cron-id lifecycle, and the guaranteed respawn (helper referenced by every mode, both holes closed, reconcile recovery). `verify:contracts` passes.

## Open design questions for review

1. **Reminder destination.** The one-line reminder goes to the **loop owner in the session** (cheapest, no GitHub notification spam), never to the PR. A sticky, edit-in-place PR comment is an alternative if the owner isn't watching the session — worth it, or is session output right?
2. **Recovery staleness window.** Reconcile re-arms an open slot whose newest `followup.log` tick is older than 15 minutes — comfortably beyond the `5m` blocked cadence. Is 15m the right margin, or too eager / too slow?
3. **Residual orphan.** A cron that survived a compaction (blind to `CronList`) whose slot was *also* deleted has no id to delete and no slot to re-arm — only a session restart clears it. The durable id plus the Concern C staleness sweep shrink this to a slot removed before its first tick. Acceptable, or warn for manual restart?
4. **Model tier.** The watcher runs on `haiku`. The fingerprint compare (Step 2.5), the blocked-vs-transient classification (Step 7), and now the cron swap are mechanical branches, so haiku should hold — flagging in case review disagrees.


stan4git:
I want to emphasis the importance of "responsiveness". I want to balance between "token cost" and "user experience": keep reminding the owner of the next act until it is answered, so do no stop or slow the poll, but rather keep the work active (poll), while in each poll provide reference (a one line summary) for user to decide next step/trace back to the decision context.

stan4git:
Reworked to prioritize responsiveness in c338dcc — the watcher no longer parks or slows on a human-blocked PR. It stays on the `1m` poll, and each blocked tick emits a **one-line reminder** to the owner: the pending act plus a reference back to the decision context (the review, the checks, or the blocked SHA), nudging until you answer. Cost stays bounded because each blocked tick is a cheap fingerprint check + one line, not a full re-evaluation; the block clears the instant a push, review, or CI/deploy state moves.

What changed vs the park version:
- `park` → `blocked` in `last_seen.json` + telemetry (`blocked`/`reminded`); no cadence swap anywhere — `cron.json.interval` is always `1m`.
- Step 7 flags the block and reminds (keeps `1m`); Step 2.5 reminds-or-resumes (falls through to a normal eval the same tick the fingerprint moves).
- New `output-templates/blocked-reminder.md` (per-reason one-liner with a traceback ref); drift-lint now asserts the blocked path never reintroduces a backoff.
- Concern B (cron-id lifecycle) unchanged.

Two follow-ups I'd value your call on (also in the PR body): **(1)** the reminder goes to the session owner, not the PR — want a sticky, edit-in-place PR comment instead for when you're away from the session? **(2)** re-emit the line every `1m` tick, or throttle the wording while keeping the `1m` fingerprint check?

plugin/skills/muggle-pr-followup/contract.md
stan4git, line 1: for paths/instruction details that is not applicable to all invocation (conditional paths), move details out into spearate files and reference here.

stan4git: Scan this whole PR and apply the same rule

stan4git: Addressed in 3683750: extracted the conditional blocked-path detail — the fingerprint, the flag-and-remind (Step 7), and the remind-or-resume gate (Step 2.5) — out of `contract.md` into a new referenced `blocked-tick.md`. `contract.md` Steps 2.5 / 7 now stay lean pointers, so only the always-runs tick flow lives inline; the blocked path (which fires only while `last_seen.blocked` is set) sits in its own file next to `cancel-cron.md` / `record-cron-id.md`.

Scanned the rest of the PR per your second note: the other changed files are schema/reference docs (`state-schemas.md`, the output-templates, the telemetry event) rather than per-invocation procedure with conditional branches, so `contract.md` was the one file carrying inlined conditional detail.


plugin/skills/_shared/telemetry-events/pr-followup-tick.md
```diff
@@ -29,3 +31,5 @@ One per watcher iteration (idle or not).
 - `dispatched_ci_fix`: true when this tick dispatched `/muggle-do` with a fix-ci directive.
 - `terminal`: true when this tick observed the PR merged or closed and wrote `result.md`.
 - `idle`: true when nothing was dispatched this tick.
+- `blocked`: true on a tick that idled on a durable human-block (escalated rebase/CI, or an ambiguous review awaiting the user) and emitted the one-line owner reminder. `false` on the tick the block clears and on every normal tick. Always accompanies `idle: true`. The cadence stays `1m` throughout — the watcher reminds rather than backing off.
```
stan4git: let's slow it down to 5m poll if blocked

stan4git: Done in 4eff8e0 — the blocked path now backs off to a 5m poll instead of staying at 1m.

- The watcher swaps its cron 1m→5m when it flags a human-block, and swaps back to 1m the instant the fingerprint moves — a real cancel-then-create, so the slot is never left cron-less. It slows to remind, never stops.
- `cron.json.interval` and the `tick` event now carry `"1m" | "5m"`, so the live cadence is observable in telemetry (this file).

Wired through `blocked-tick.md` (the new Cadence-swap section), `contract.md` (Steps 2.5/7), `state-schemas.md`, and the reminder/log templates. Drift-lint updated to assert the 5m backoff instead of the old always-1m rule.

plugin/skills/muggle-pr-followup/cancel-cron.md
```diff
@@ -1,11 +1,12 @@
 # Cancel the watcher's cron
 
-The find-and-delete every tick uses to stop its own loop: the stale-fire guard and terminal unschedule ([`finalize.md`](finalize.md) Step 4), and each single-thread "stop this watcher" before a `/muggle-do` dispatch ([`contract.md`](contract.md) Steps 4–6).
+The delete every tick uses to stop its own loop: the stale-fire guard and terminal unschedule ([`finalize.md`](finalize.md) Step 4), and each single-thread "stop this watcher" before a `/muggle-do` dispatch ([`contract.md`](contract.md) Steps 4–6). The blocked path also uses this delete as the cancel half of its cadence swap — cancel the current cron, then `CronCreate` the same command at the target interval ([`blocked-tick.md`](blocked-tick.md)) — so a blocked watcher slows to `5m` and restores `1m` without ever leaving the slot cron-less ([`contract.md`](contract.md) Steps 2.5, 7).
```
stan4git: why do you need to introduce reverse dependency?

This cancel cron just do cancel cron, let the caller worry about what is after

stan4git: Fixed in 48970f8 — you're right, that was a reverse dependency. `cancel-cron.md` now documents only the delete ("deletes this slot's watcher cron, and nothing more — the caller owns whatever comes next") and just names its callers, without describing the swap or linking up. The blocked-path cadence swap lives entirely in `blocked-tick.md`, which links *down* to cancel-cron for the cancel half — the one correct direction.

plugin/skills/_shared/telemetry-events/pr-followup-tick.md
```diff
@@ -29,3 +32,6 @@ One per watcher iteration (idle or not).
 - `dispatched_ci_fix`: true when this tick dispatched `/muggle-do` with a fix-ci directive.
 - `terminal`: true when this tick observed the PR merged or closed and wrote `result.md`.
 - `idle`: true when nothing was dispatched this tick.
+- `blocked`: true on a tick that idled on a durable human-block (escalated rebase/CI, or an ambiguous review awaiting the user) and emitted the one-line owner reminder. `false` on the tick the block clears and on every normal tick. Always accompanies `idle: true`. While blocked the cadence backs off to `5m` (`interval: "5m"`); it returns to `1m` the moment the block clears — the watcher slows to remind, never stops.
```
stan4git: what is the specific case this state provide any value? only change of cadence?

Feels off. you could have just use one state "blocked_on_user_reply" for this case? instead of using blocked + reminded. reminded not providing any extra value since you does it ever tick?

On the other hand, let's keep the poll interval 1m consistently (haiku not going to cost much) so user can get timely response

stan4git: Addressed in 649bbf4 — you're right on all three:

- **1m always.** Removed the 5m backoff and the whole cadence-swap machinery. The poll stays at 1m whether or not the PR is blocked; haiku is cheap and the owner gets a timely nudge. It reminds but never slows or stops.
- **Dropped `reminded`.** It was `true` on every `blocked: true` tick by definition, so it carried no information. Also dropped `interval` from the event (it's a constant 1m now).
- **One state.** Kept `blocked` as the single state.

On *"what value does `blocked` provide?"* — now that it changes no cadence, its only job is (a) the reason-specific one-line owner reminder and (b) fingerprint-based **auto-resume** (head SHA / latest review id / CI digest): the watcher clears the block itself the instant a push, review, or CI/deploy change lands, instead of the owner having to poke it. It also caches `since` for "blocked for 2h" wording.

So it's a small cached flag that drives the reminder + auto-resume. If you'd rather drop it entirely and recompute the escalation state inline each tick (no cached `blocked` object at all), I can — that's a slightly bigger change, so say the word and I'll do it.


plugin/skills/_shared/telemetry-events/pr-followup-tick.md
```diff
@@ -29,3 +30,4 @@ One per watcher iteration (idle or not).
 - `dispatched_ci_fix`: true when this tick dispatched `/muggle-do` with a fix-ci directive.
 - `terminal`: true when this tick observed the PR merged or closed and wrote `result.md`.
 - `idle`: true when nothing was dispatched this tick.
+- `blocked`: true on a tick that idled on a durable human-block (escalated rebase/CI, or an ambiguous review awaiting the user) and emitted the one-line owner reminder. `false` on the tick the block clears and on every normal tick. Always accompanies `idle: true`. The cadence stays `1m` whether blocked or not — a blocked tick reminds at `1m`, it does not back off. A `blocked: true` tick always emits exactly one owner reminder (the reminder is implied by the flag, so no separate `reminded` field is carried).
```
stan4git: keep it tight

stan4git: Tightened in 1366d5c — trimmed the `blocked` note here (and the parallel notes in `state-schemas.md` and the `blocked-tick.md` governing rule) that I'd bloated in 649bbf4 explaining the new design. Same meaning, fewer words.

PR #316, opened 2026-07-10
feat(skills): enforce one-way skill dependencies with a deterministic gate

stan4git:
## Why

`plugin/skills/CLAUDE.md` documents a **one-way dependency rule** — if skill A links to B, no file in B may link back to A — but nothing enforced it. Reverse dependencies kept getting introduced. This makes the rule a **deterministic gate** instead of a hope, **and cleans up every violation it found** so the tree ships genuinely one-way.

## What a "reference" is

Exactly what the rule says: a **markdown file-link into another skill's directory**. Prose mentions, descriptions naming other skills, and runtime `/slash-command` dispatch are **not** references and are ignored — so the gate doesn't false-positive on the ~300 legitimate cross-skill links.

## How it works

`scripts/check-skill-deps.mjs` (zero-dep Node) derives the cross-skill link graph and fails on any **cycle** — a cycle *is* a reverse dependency (A→B and B→A). It runs three ways:

| Surface | Catches |
|---|---|
| `skill-deps` CI job | reverse deps in **any** agent's PR, at merge — the unconditional backstop |
| `PreToolUse` deny hook (`.claude/settings.json`) | the write in-session, naming the offending link, before it lands |
| `pnpm run verify:skill-deps` | local check |

Modeling notes (in `plugin/skills/skill-deps.config.json`):
- `_shared` is a namespace of independent modules, not one skill → **exploded to per-file nodes** so unrelated shared modules don't manufacture false cycles.
- `do/` is `muggle-do`'s externalized step files → **grouped into `muggle-do`**.

## The reverse deps it found — now fixed

The gate caught **9 real reverse-dependency links already in the tree** — shared modules linking *up* into `muggle-do`'s `do/` steps and `muggle-pr-followup`'s templates/schemas, plus one intra-`_shared` 2-cycle. Every one is now fixed by genericizing the shared module to describe the **contract** and letting the caller own the specifics — which matches each file's own stated intent (*"the caller passes in"*, *"this file names no caller"*):

- `dev-loop/run`, `resolve-rebase-conflicts` — drop links into `muggle-do`'s `do/` verify steps; the caller runs its own build/unit/E2E
- `resolve-rebase-conflicts`, `vcs/github/verify-working-tree` — drop links into `muggle-pr-followup` templates; the caller renders its own escalation / wrong-checkout message
- `resolve-e2e-validation-context` — stop deferring question wording to `do/pre-flight`; ask the same intent, caller supplies option text
- `session-state-writes` — point at its own `## Field map`, not the watcher's `state-schemas`
- `use-worktrees` — stop linking down to the PR-branch specialization (the general file must not point at the specific)

**`knownReverseDeps` is now empty.** The tree is genuinely one-way: 307 cross-skill links, 0 cycles. No behavior change — only reverse *links* removed; the information each shared file needs is preserved (inlined, self-anchored, or owned by the caller that was already running the procedure).

## Verified

- Full lint green with an **empty baseline**: `307 cross-skill links, no reverse dependencies`
- Hook **denies** a new reverse edge (`muggle-pr-followup → muggle-do`) with a precise reason; **allows** downward links, non-skill files
- Handles Write / Edit / MultiEdit payloads; safe-allows on unparseable input
- eslint / typecheck / build / all CI green

https://claude.ai/code/session_01NLDR4ox777q3M1un5AdhbM

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -23,6 +23,18 @@
           }
         ]
       }
+    ],
+    "PreToolUse": [
+      {
+        "matcher": "Edit|Write|MultiEdit",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node \"${CLAUDE_PROJECT_DIR}/scripts/check-skill-deps.mjs\" --hook",
+            "statusMessage": "Checking skill dependency direction"
+          }
+        ]
+      }
     ]
   }
 }
--- a/plugin/skills/_shared/dev-loop/run.md
+++ b/plugin/skills/_shared/dev-loop/run.md
@@ -1,6 +1,6 @@
 # Dev Loop — Run a Test
 
-> The local dev loop: run one test case in the browser (replay an existing script, or regenerate from the case), then record the result. Source of truth for the run mechanics; the sibling files in this folder own each invariant. Used by `muggle-test`, `muggle-test-feature-local`, `muggle-do` Stage 6, and the `acceptance-tester` agent.
+> The local dev loop: run one test case in the browser (replay an existing script, or regenerate from the case), then record the result. Source of truth for the run mechanics; the sibling files in this folder own each invariant.
 
 Not owned here — the caller resolves and passes in: which test cases to run, replay-vs-regen classification and failure routing ([`../failure-mode-handling.md`](../failure-mode-handling.md)), dev-server readiness ([`../dev-server-readiness.md`](../dev-server-readiness.md)), validation context ([`../resolve-e2e-validation-context.md`](../resolve-e2e-validation-context.md)), and PR posting ([`../../muggle-pr-visual-walkthrough/SKILL.md`](../../muggle-pr-visual-walkthrough/SKILL.md)).
 
--- a/plugin/skills/_shared/dev-loop/run.md
+++ b/plugin/skills/_shared/dev-loop/run.md
@@ -1,6 +1,6 @@
 # Dev Loop — Run a Test
 
-> The local dev loop: run one test case in the browser (replay an existing script, or regenerate from the case), then record the result. Source of truth for the run mechanics; the sibling files in this folder own each invariant. Used by `muggle-test`, `muggle-test-feature-local`, `muggle-do` Stage 6 ([`../../do/e2e-acceptance.md`](../../do/e2e-acceptance.md)), and the `acceptance-tester` agent.
+> The local dev loop: run one test case in the browser (replay an existing script, or regenerate from the case), then record the result. Source of truth for the run mechanics; the sibling files in this folder own each invariant. Used by `muggle-test`, `muggle-test-feature-local`, `muggle-do` Stage 6, and the `acceptance-tester` agent.
 
 Not owned here — the caller resolves and passes in: which test cases to run, replay-vs-regen classification and failure routing ([`../failure-mode-handling.md`](../failure-mode-handling.md)), dev-server readiness ([`../dev-server-readiness.md`](../dev-server-readiness.md)), validation context ([`../resolve-e2e-validation-context.md`](../resolve-e2e-validation-context.md)), and PR posting ([`../../muggle-pr-visual-walkthrough/SKILL.md`](../../muggle-pr-visual-walkthrough/SKILL.md)).
 
--- a/plugin/skills/_shared/resolve-rebase-conflicts.md
+++ b/plugin/skills/_shared/resolve-rebase-conflicts.md
@@ -1,14 +1,13 @@
-# Auto-Resolve Rebase Conflicts
+# Resolve Rebase Conflicts
 
-The autonomous conflict-resolution body, run when a rebase onto `origin/{default}` reports conflicts **and** [`autoResolveConflicts`](../muggle-preferences/preference-gates/autoResolveConflicts.md) is `always`. Under `never` the caller aborts and escalates instead, and this file never runs. The caller hands off the inputs below; this file names no caller — the dependency runs one way.
+How to resolve the conflicts a rebase onto `origin/{default}` reports, when [`autoResolveConflicts`](../muggle-preferences/preference-gates/autoResolveConflicts.md) is `always`. Under `never` the caller aborts and escalates instead, and this file never runs. The caller hands off the inputs below and, once resolution completes, runs the [`verify-or-rollback-gate.md`](verify-or-rollback-gate.md) before anything ships. This file names no caller — the dependency runs one way.
 
-Contract: never push an auto-resolved rebase that has not passed the verify gate, and always keep the branch restorable to its pre-rebase state.
+Contract: resolve deterministically where the change is mechanical, reason about intent where it is semantic, and never guess in load-bearing logic — hand an unreconcilable conflict back to the caller rather than fabricating a merge.
 
 ## Inputs
 
-- `pre_rebase_sha` — branch HEAD captured by the caller **before** `git rebase`; the rollback point.
+- `pre_rebase_sha` — branch HEAD captured by the caller **before** `git rebase`; the caller passes it on to the verify-or-rollback gate as the rollback point.
 - `default` — the branch being rebased onto.
-- Session context: slug, PR url/number, and the persisted validation strategy (for the E2E step).
 
 ## Procedure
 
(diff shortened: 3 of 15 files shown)


plugin/skills/skill-deps.config.json
```diff
@@ -0,0 +1,18 @@
+{
+  "supportDirs": { "do": "muggle-do" },
+  "sharedNamespaces": ["_shared"],
+  "knownReverseDeps": {
+    "comment": "Pre-existing reverse dependencies grandfathered when the one-way guard was added. Each is a genuine violation of plugin/skills/CLAUDE.md (a shared module or lower-level skill linking UP into a caller) — debt to fix and delete from this list, not a pattern to copy. The guard blocks any NEW cycle regardless of this list; these entries only keep the pre-existing tangle from failing CI. Burn this list down to empty.",
+    "edges": [
```
stan4git: how about create a reference section to map a name to the path, and use the name here in the graph?

stan4git: Good call — but it turns out to be moot at the root: the follow-up commit (18c5afe) fixed all 9 reverse deps, so `knownReverseDeps.edges` is now empty. There are no path references left in the config to name.

If we ever need to re-grandfather an exception, I agree — I'd key it by node name (e.g. `"to": "muggle-do"`, which is what the detector already resolves targets to) rather than a raw relative path like `../../do/e2e-acceptance.md`. But shipping the name→path legend now would abstract over zero entries.

Want it pre-built for the future, or added when the first real entry appears?


.claude/settings.json
```diff
@@ -23,6 +23,18 @@
...
         ]
       }
+    ],
+    "PreToolUse": [
+      {
+        "matcher": "Edit|Write|MultiEdit",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node \"${CLAUDE_PROJECT_DIR}/scripts/check-skill-deps.mjs\" --hook",
```
stan4git, line 33: How much perf drag as a hook vs CI?

stan4git: Measured it — the graph analysis isn't the cost; the Node process spawn is.

**Pure scan work:** reading + link-parsing all 170 skill `.md` files is **~7 ms per full-tree pass** (20 passes in one process = 134 ms). The hook's two-pass baseline-vs-proposed diff is therefore ~13 ms of real work; the Tarjan SCC pass over a few hundred edges is noise.

**So the only real drag is one `node` cold-start:**
- **CI** — one spawn + one ~7 ms scan **per PR**. Off the hot path, invisible.
- **Hook** — one spawn **per `Edit`/`Write`/`MultiEdit`**, but the script early-exits after the stdin read + a `.md`/skill-path check (`check-skill-deps.mjs` lines 284–289):
  - non-`.md` or non-skill edits (the overwhelming majority): ~0 scan work — you pay only the Node cold-start (~50–150 ms on an unloaded box).
  - skill-`.md` edits (rare): + the ~13 ms two-pass scan, lost in the spawn noise.

Net: the hook adds roughly one Node cold-start per edit, dominated by process spawn rather than this script. Sub-150 ms on a normal dev box, invisible interactively; scan cost grows ~linearly with skill-`.md` count (170 files → 7 ms), so lots of headroom.

If the per-edit spawn ever matters, the matcher (`Edit|Write|MultiEdit`) can be scoped to skill paths so Node isn't spawned on unrelated edits at all — happy to do that as a follow-up. Keeping it broad + self-filtering is what lets the hook block a reverse-dep the moment it's written mid-session, which CI can only catch post-push.

plugin/skills/_shared/resolve-rebase-conflicts.md
stan4git: I feel this gate should goes into the callers (build, etc). 

Let this file be how to resolve only

plugin/skills/_shared/use-worktrees.md
```diff
@@ -7,7 +7,7 @@ Use this for feature development, local validation, and PR iteration.
 - **One worktree per branch.** Never switch branches inside a long-lived checkout.
 - **Isolate runtime resources.** Parallel worktrees need unique ports and isolated mutable test state.
 - **Keep worktrees disposable.** Create for focused work, remove after merge.
-- **Existing PR branch?** Materialize it via [`pr-branch-worktree.md`](pr-branch-worktree.md) — same `<repo>/.claude/worktrees/` path scheme, checking out the existing branch instead of creating a new one.
+- **Existing PR branch?** Check it out into the same `<repo>/.claude/worktrees/` path scheme instead of creating a new branch. The skill that takes a PR URL owns that operation.
```
stan4git: I think this should belong to the parent md

stan4git: Agreed — moved. That bullet described the materialize-an-existing-PR-branch operation, which `pr-branch-worktree.md` already owns end-to-end (it builds the same `<repo>/.claude/worktrees/<branch>` path and checks out the existing branch rather than creating one, and its header says *"Skills MUST link here rather than restate the steps"*).

Since `use-worktrees.md` is the low-level shared primitive, keeping a bullet here that reaches up toward *"the skill that takes a PR URL"* is exactly the upward coupling this PR's guard discourages. Removed it from `use-worktrees.md` in 48a08ce; the parent `pr-branch-worktree.md` remains the single source of truth. Guard still green (310 links, no reverse deps).

plugin/skills/_shared/resolve-rebase-conflicts.md
stan4git: Good separation-of-concerns instinct — but this one needs a design call before I refactor, so flagging rather than guessing.

The verify gate (Steps 3–5) isn't invoked by `build`; its two actual callers are `_shared/rebase-before-e2e.md` and `do/resolve-conflicts.md`, and both currently *delegate* the gate down to this file. A literal "move it into the callers" would duplicate the build → unit → E2E → rollback-on-fail sequence into **both** callers — which runs into the repo's single-source-of-truth rule. It's also the file's stated Contract invariant ("never push an auto-resolved rebase that has not passed the verify gate"), so wherever it lands it has to stay mandatory.

**Recommendation:** extract Steps 3–5 into a new `_shared/verify-or-rollback-gate.md` that both callers invoke after a resolve, and reduce this file to how-to-resolve only (Steps 1–2 + the rollback mechanism it hands back). That honors "let this file be how to resolve only" *and* keeps the gate DRY across the two callers.

Confirm that direction (or tell me you'd rather each caller own its own copy) and I'll do the refactor next tick. Holding off on code until you pick.

stan4git:
These threads are addressed and still open — mark them resolved if satisfied, or reply if more is needed:
- [`_shared/use-worktrees.md` — PR-branch bullet](https://github.com/multiplex-ai/muggle-ai-works/pull/316#discussion_r3567385665) (moved; parent `pr-branch-worktree.md` owns it)
- [`skill-deps.config.json` — knownReverseDeps comment](https://github.com/multiplex-ai/muggle-ai-works/pull/316#discussion_r3562217571)

(The `resolve-rebase-conflicts.md` gate thread is left open on purpose — it's awaiting your design call, not resolution.)


plugin/skills/_shared/resolve-rebase-conflicts.md
stan4git: looks good to me. Do it

stan4git: Done in 6d67737.

- New `_shared/verify-or-rollback-gate.md` holds the old Steps 3–5 (verify → pass/proceed → fail/restore+escalate). Dispatch-free, names no caller.
- `resolve-rebase-conflicts.md` is now enumerate → classify → resolve only; on an unreconcilable conflict it hands back to the caller, which restores `pre_rebase_sha` and escalates via the gate.
- `rebase-before-e2e.md` resolves then runs the gate. `resolve-conflicts.md`'s own Steps 3/5 are its instance of the gate (they carry the extra attempt-budget + force-push logic), so it points at the shared file rather than duplicating it.

Mandatory-verify invariant preserved; skill-dep guard green (314 links, no reverse deps).

stan4git:
This thread is addressed and still open — mark it resolved if satisfied, or reply if more is needed:
- [`_shared/resolve-rebase-conflicts.md` — verify-or-rollback gate extraction](https://github.com/multiplex-ai/muggle-ai-works/pull/316#discussion_r3567382955) (done in 6d67737)


plugin/skills/_shared/dev-loop/run.md
```diff
@@ -1,6 +1,6 @@
 # Dev Loop — Run a Test
 
-> The local dev loop: run one test case in the browser (replay an existing script, or regenerate from the case), then record the result. Source of truth for the run mechanics; the sibling files in this folder own each invariant. Used by `muggle-test`, `muggle-test-feature-local`, `muggle-do` Stage 6 ([`../../do/e2e-acceptance.md`](../../do/e2e-acceptance.md)), and the `acceptance-tester` agent.
+> The local dev loop: run one test case in the browser (replay an existing script, or regenerate from the case), then record the result. Source of truth for the run mechanics; the sibling files in this folder own each invariant. Used by `muggle-test`, `muggle-test-feature-local`, `muggle-do` Stage 6, and the `acceptance-tester` agent.
```
stan4git: why you reference this md's callers again ???

plugin/skills/_shared/session-state-writes.md
```diff
@@ -1,6 +1,6 @@
 # Writing session-state files
 
-How to update the session-state JSON — `last_seen.json` and `prs.json` under `~/.muggle-ai/muggle-do/sessions/<slug>/`. Field shapes: [`../muggle-pr-followup/state-schemas.md`](../muggle-pr-followup/state-schemas.md). Both the watcher and `/muggle-do` write these.
+How to update the session-state JSON — `last_seen.json` and `prs.json` under `~/.muggle-ai/muggle-do/sessions/<slug>/`. Field shapes: see the [`## Field map`](#field-map) below. Both the watcher and `/muggle-do` write these.
```
stan4git: why call out the parent again here????

stan4git:
Re: [`_shared/dev-loop/run.md:3`](https://github.com/multiplex-ai/muggle-ai-works/pull/316#discussion_r3568466509) — *"why you reference this md's callers again"* (replying top-level: you have a pending review open, so a threaded reply from this account 422s).

Fair — dropped it in d2204aa. A shared primitive listing its own callers is the same upward coupling the guard is here to kill; the header now stops at "the sibling files in this folder own each invariant," and the "what the caller passes in" section below already points the dependency the right way (downward, into siblings).

On the "again": the identical `Used by <caller list>` pattern also sits in three **pre-existing** `_shared` headers this PR doesn't otherwise touch —

- `pr-branch-worktree.md:3` — "Used by `muggle-test` …"
- `failure-mode-handling.md:3` — "… used by `muggle-test`, `muggle-test-feature-local`, `muggle-browser-task`, `muggle-test-regenerate-missing`"
- `debug-failed-run.md:3` — "Used by `muggle-test` (Step 7C), `muggle-test-feature-local`, and the `muggle-do` e2e stage"

I left them out to keep this PR scoped to what the guard changed. Say the word and I'll sweep all three (here or a follow-up) — your call on scope.

plugin/skills/_shared/session-state-writes.md
stan4git: Same fix — removed in 9772ad0. "Both the watcher and `/muggle-do` write these" was the shared file naming its callers; the header now stops at the field-map pointer. That's the last of the `_shared` headers in this PR's diff that called out their callers (use-worktrees, dev-loop/run, resolve-rebase-conflicts, now this).

PR #320, opened 2026-07-19
feat(pr-followup): auto-reconcile stale watchers at session start

stan4git:
## Gap

`muggle-pr-followup` watchers are session-only `/loop` crons — they die on session end and after the 7-day `/loop` expiry. The skill already ships a recovery procedure, `reconcile.md`: it finalizes slots whose PR went terminal while polling lapsed, sweeps orphan crons, and re-arms open slots whose watcher stopped silently.

But reconcile had **no automatic trigger** — it ran only on a manual `/muggle:muggle-pr-followup reconcile` or a no-arg auto-track invocation. So open PRs quietly accumulated permanently-dead watchers with nothing to notice. Real evidence: 9 open PRs sat un-watched for ~7 weeks.

## Fix

A `SessionStart` hook can't invoke a skill and can't call `CronCreate` to re-arm, so the trigger is a **nudge**, not an action:

`plugin/scripts/reconcile-stale-watchers.sh` scans `~/.muggle-ai/muggle-do/sessions/*/` for **open slots** (a `prs.json` with no `result.md`). When one or more exist, it injects `additionalContext` (the same `SessionStart` contract the existing `ensure-electron-app.sh` hook uses) telling the agent to run `/muggle:muggle-pr-followup reconcile`. Reconcile then does the real work — finalizing terminal slots and re-arming genuinely-dead open watchers.

It's a pure directory scan under `${HOME}` — no `gh`, no network, no writes — cheap enough to run on every session start.

## Silent when there's nothing to do

Zero open slots → the script emits nothing and exits 0. It only ever nudges when a non-terminal slot actually exists, and reconcile itself is idempotent (re-arms only stale watchers, leaves live ones alone), so a clean workspace produces no noise.

## Changes

- **`plugin/scripts/reconcile-stale-watchers.sh`** (new) — the scan-and-nudge hook script.
- **`plugin/hooks/hooks.json`** — registers it as a second `SessionStart` command (matcher `startup|clear|compact`, `timeout: 10`), alongside the existing `ensure-electron-app.sh`.
- **`reconcile.md` / `SKILL.md` / `CLAUDE.md` / `hooks/README.md`** — document the new session-start trigger; the recover-don't-seed invariant is unchanged (a session-start run still never seeds a first watcher).
- **`src/test/skills/pr-followup-reconcile-hook.test.ts`** (new) — drift-lint: the hook is registered and points at the script; the script keys open-slot on `prs.json` present + `result.md` absent, its empty-state exit precedes any emission, it stays a pure scan, and the docs carry the trigger.

## Verification

- `npx vitest run` — 677 passed, 5 skipped (full suite).
- `node scripts/check-skill-deps.mjs` — OK, no reverse dependencies.
- Functional smoke of the script: 11 real open slots → nudge emitted; clean `HOME` and all-terminal slots → silent (exit 0, no output); singular/plural wording correct.

### DIFF
--- a/plugin/skills/muggle-pr-followup/SKILL.md
+++ b/plugin/skills/muggle-pr-followup/SKILL.md
@@ -35,6 +35,8 @@ The skill recognizes its mode by inspecting `$ARGUMENTS` and falling back to on-
 
 Auto-track runs **reconcile** first, so a no-arg invocation also finalizes any slot whose PR merged or closed while its watcher was down (expired cron, ended session) and re-arms any open slot whose watcher stopped silently (a dropped respawn). Reconcile recovers a watcher that was already running; it never seeds a first watcher for a PR — that is auto-track's / bootstrap's job.
 
+**Reconcile is also triggered at session start**, so a dead watcher gets caught without waiting for a manual sweep: a `SessionStart` hook ([`../../hooks/README.md`](../../hooks/README.md)) scans for open slots and, only when some exist, nudges the agent to run reconcile. Session-only crons die on session end and the 7-day `/loop` expiry, so without this a merged-away PR's slot could sit un-finalized indefinitely. The hook only nudges — it can't call `CronCreate` to re-arm — and stays silent when no open slot exists. See [`reconcile.md`](reconcile.md#triggers).
+
 Bootstrap accepts three optional trailing flags:
 
 - `--slug=<name>` — override the default `<repo>-pr<n>` slug
--- a/plugin/skills/muggle-pr-followup/SKILL.md
+++ b/plugin/skills/muggle-pr-followup/SKILL.md
@@ -35,7 +35,7 @@ The skill recognizes its mode by inspecting `$ARGUMENTS` and falling back to on-
 
 Auto-track runs **reconcile** first, so a no-arg invocation also finalizes any slot whose PR merged or closed while its watcher was down (expired cron, ended session) and re-arms any open slot whose watcher stopped silently (a dropped respawn). Reconcile recovers a watcher that was already running; it never seeds a first watcher for a PR — that is auto-track's / bootstrap's job.
 
-**Reconcile is also triggered at session start**, so a dead watcher gets caught without waiting for a manual sweep: a `SessionStart` hook ([`../../hooks/README.md`](../../hooks/README.md)) scans for open slots and, only when some exist, nudges the agent to run reconcile. Session-only crons die on session end and the 7-day `/loop` expiry, so without this a merged-away PR's slot could sit un-finalized indefinitely. The hook only nudges — it can't call `CronCreate` to re-arm — and stays silent when no open slot exists. See [`reconcile.md`](reconcile.md#triggers).
+**Reconcile also runs at session start** — a `SessionStart` hook ([`../../hooks/README.md`](../../hooks/README.md)) surfaces the sweep when open slots exist, catching a watcher that died with its session (end, or 7-day `/loop` expiry) before a manual sweep would. See [`reconcile.md`](reconcile.md#triggers).
 
 Bootstrap accepts three optional trailing flags:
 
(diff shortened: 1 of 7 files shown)


plugin/skills/muggle-pr-followup/reconcile.md
```diff
@@ -4,6 +4,16 @@ The procedure for the **reconcile mode** of `muggle-pr-followup` — a sweep tha
 
 Termination is otherwise tick-driven ([`contract.md`](contract.md) Step 2): a slot finalizes only when a tick fires and observes `MERGED` / `CLOSED`. If the tick stream stops first — the recurring `/loop` cron auto-expires after 7 days, the session ends, or the machine is off when the PR merges — no tick catches the transition, and the slot is left un-finalized: no `result.md`, no post-merge cleanup, and a surviving cron would keep polling a dead PR. Reconcile is the catch-up.
 
+## Triggers
```
stan4git, line 7: WHY describe the caller's behaviors? reverse dep isn't it?

plugin/skills/muggle-pr-followup/SKILL.md
```diff
@@ -35,6 +35,8 @@ The skill recognizes its mode by inspecting `$ARGUMENTS` and falling back to on-
 
 Auto-track runs **reconcile** first, so a no-arg invocation also finalizes any slot whose PR merged or closed while its watcher was down (expired cron, ended session) and re-arms any open slot whose watcher stopped silently (a dropped respawn). Reconcile recovers a watcher that was already running; it never seeds a first watcher for a PR — that is auto-track's / bootstrap's job.
 
+**Reconcile is also triggered at session start**, so a dead watcher gets caught without waiting for a manual sweep: a `SessionStart` hook ([`../../hooks/README.md`](../../hooks/README.md)) scans for open slots and, only when some exist, nudges the agent to run reconcile. Session-only crons die on session end and the 7-day `/loop` expiry, so without this a merged-away PR's slot could sit un-finalized indefinitely. The hook only nudges — it can't call `CronCreate` to re-arm — and stays silent when no open slot exists. See [`reconcile.md`](reconcile.md#triggers).
```
stan4git: tight this up.

plugin/skills/muggle-pr-followup/reconcile.md
stan4git: Addressed in 12869fc: dropped the hook's internals (scan / nudge-only / can't-re-arm) from the session-start trigger — that was the reverse dependency. It now just names `reconcile-stale-watchers.sh` as the trigger and links `hooks/README.md` for the behavior.


plugin/skills/muggle-pr-followup/SKILL.md
stan4git: Addressed in 12869fc: tightened the session-start paragraph to a single line — trigger + why + a downward link to `reconcile.md#triggers`.


stan4git:
Addressed the two review threads in `12869fc` — mark them resolved when satisfied:
- `reconcile.md` — dropped the hook's internals from the session-start trigger (the reverse dependency)
- `SKILL.md` — tightened the session-start paragraph


🤖 _Automated resolve reminder from muggle-do._

plugin/skills/muggle-pr-followup/reconcile.md
stan4git, line 7: I don't understand, why "reconcile" needs to know who/how triggered it

plugin/skills/muggle-pr-followup/reconcile.md
```diff
@@ -4,6 +4,16 @@ The procedure for the **reconcile mode** of `muggle-pr-followup` — a sweep tha
...
+
+Three ways in, all running the same procedure:
+
+- **Manual** — `/muggle:muggle-pr-followup reconcile` (or `sweep`).
+- **Auto-track** — the top of a no-arg invocation ([`auto-track.md`](auto-track.md)).
+- **Session start** — the `reconcile-stale-watchers.sh` hook ([`../../hooks/README.md`](../../hooks/README.md)) runs this sweep, catching a watcher that died with its session (end, or 7-day `/loop` cron expiry) before its PR's merge was observed.
```
stan4git, line 13: do you need to update the dep graph?

stan4git: No — verified. `check-skill-deps` passes (314 links, no reverse dependencies). The edit only removed prose and kept a downward link to `hooks/README.md`; hooks aren't skills, so no skill-graph edge changed.


plugin/skills/muggle-pr-followup/reconcile.md
stan4git: Mostly it doesn't. The manual / auto-track entries are within-skill invocation modes, but the **session-start** entry names an external caller (`reconcile-stale-watchers.sh`) — the reverse reference you're flagging. It's still named only because `pr-followup-reconcile-hook.test.ts:121` asserts reconcile.md mentions the script.

Proposed: drop the session-start trigger from reconcile.md entirely (its wiring belongs in SKILL.md + hooks/README.md — the caller's altitude) and relax that one test assertion. Holding until you confirm, since it edits a test this PR added.

PR #329, opened 2026-07-22
refactor(pr-followup): poll PRs from a haiku-pinned watcher agent

stan4git:
Idle PR polling used to run as per-minute turns inside the session that armed the watcher, re-reading its whole transcript each tick and inheriting its model (the skill's `model: haiku` pin never applied). Measured: ~190M tokens / ~$207 per 30 min across 7 watchers, >99% of it cache reads.

## Design

**Arming = drain, then watch** (`muggle-pr-followup/arm-watcher.md`, shared by bootstrap, auto-track, and the executor's respawn):

1. One tick per `contract.md` drains everything already actionable — existing threads, body-only reviews, stale branch, red CI, terminal state.
2. If nothing dispatched, spawn the `pr-watcher` agent (`plugin/agents/pr-watcher.md`, pinned `model: haiku`). It polls in its own context about every 60s and returns on new comment activity (new review id, new thread comment, or a newly unresolved thread), terminal PR, or fetch failure.
3. On return, the orchestrator runs the tick again. All detection semantics stay in `contract.md`; the poller only change-detects and never classifies.

Idle watchers cost the session zero turns and zero context. The drain step keeps the documented promise that a PR bootstrapped with existing feedback is picked up on the first tick. Crons remain only as `reconcile.md`'s recovery substrate.

## Verification

- `vitest run src/test/agents/ src/test/skills/` — 289 passed
- `node scripts/check-skill-deps.mjs` — no reverse dependencies
- Static lint on the agent file enforces the haiku pin, both detection sources, shell/OS agnosticism, and bans restating any muggle-pr-followup decision rule or state field


plugin/skills/muggle-pr-followup/bootstrap.md
```diff
@@ -83,11 +83,7 @@ Create `iterations/` subdir (empty) for future caller use.
 
 ### Step 8 — Dispatch the first watcher
 
-The last action of this turn:
-
-```
-/loop 1m /muggle:muggle-pr-followup <slug> <n>
-```
+The last action of this turn: spawn the [`pr-watcher`](../../agents/pr-watcher.md) agent with the slug, repo, and PR number. It owns the `1m` poll inside its own context and returns only when a new comment lands or the PR goes terminal; act on its decision by running [`contract.md`](contract.md). This replaces `/loop 1m /muggle:muggle-pr-followup <slug> <n>`, which billed the whole session's context every idle minute. The cron path stays as the recovery substrate ([`reconcile.md`](reconcile.md)), so `cron.json` is still seeded in Step 7.
```
stan4git: why do you babble about history cmd??? the fuck??

plugin/agents/pr-watcher.md
stan4git, line 1: why this whole shit is not tool agnostic?? not os agnostic??

plugin/skills/muggle-pr-followup/bootstrap.md
stan4git: Cut it — that sentence described what Step 8 used to be, not what it does. Also removed the stale `/loop` references in Step 9, in the `cron.json` seeding note, and in the success-summary template; the remaining `/loop` mentions describe the cron recovery substrate and are still accurate.


plugin/agents/pr-watcher.md
stan4git: Dropped the `tools: Bash` pin and reworded the loop off shell idioms; the linked recipes are now named as the only provider contract, to be issued with whatever command tool the host offers. Tradeoff: the pin was also what structurally stopped the agent reading `last_seen.json`, so that guard is now prose plus a test asserting the prose.


plugin/agents/pr-watcher.md
```diff
@@ -0,0 +1,54 @@
+---
+name: pr-watcher
+description: "PR comment poller — watches one pull request in its own context and returns the moment new comment activity lands or the PR goes terminal. Detection only: it does not classify feedback, run the dev cycle, or post anything. Spawned by muggle-pr-followup's arming procedure so idle polling never touches the orchestrating session."
```
stan4git: tight up

plugin/agents/pr-watcher.md
stan4git, line 1: Where do you maintain the state of poller?

Say if the poller died, and user called pr-followup skill again, what will happen?

plugin/agents/pr-watcher.md
```diff
@@ -0,0 +1,54 @@
+---
+name: pr-watcher
+description: "Watches one PR in its own context; reports on new comment activity or terminal state. Detection only — never classifies, executes, or posts. Spawned and resumed by muggle-pr-followup's arming procedure."
```
stan4git: why do you need to name the caller here

plugin/agents/pr-watcher.md
stan4git: Tightened — 40 words down from 60.

plugin/agents/pr-watcher.md
stan4git: Nowhere durable — the poller holds only an in-context baseline, re-derived from GitHub each time it wakes; all durable state (watermark, budgets, escalations) stays in the session slot. If the poller dies and you call pr-followup again, bootstrap --resume re-runs arming: drain tick handles anything missed, then a fresh watcher spawns — nothing is lost because nothing durable lived in it. Encoded in arm-watcher.md step 2.

plugin/agents/pr-watcher.md
stan4git: No reason — dropped. The description now says only what it does; who arms it is the caller's business.

src/test/agents/pr-watcher-agent.test.ts
stan4git, line 1: can you please add test between skill and arming of this agent?

stan4git: Added — three wiring assertions: arm-watcher.md spawns this agent and names it PR #<n> — <title>; the drain tick precedes the spawn (index order); and bootstrap, auto-track, and respawn-watcher all arm through arm-watcher.md.

PR #331, opened 2026-07-23
refactor(pr-followup): watch runs as a visible persistent monitor, gone only at terminal

stan4git:
Follow-up to #329, from live dogfooding feedback: the agent-based watch showed 'completed' in the UI while its background loop polled — the watch was invisible exactly while it was working. Agents have no visible-idle state, and foreground waits auto-background at 120s, so no agent substrate can satisfy 'explicitly showing while polling.'

## Design

The watch is now a **persistent labeled monitor** in the orchestrating session (`arm-watcher.md`): visible as a running task named `PR #<n> — <title>` for as long as the PR is polled, and gone exactly when it reports, the PR terminates, or fetches persistently fail. Zero model tokens while quiet. The drain-then-watch sequence, PR-derived naming, re-arm-after-every-cycle rule, and watermark-passed-in requirement are unchanged; the `pr-watcher` agent definition and its test are deleted — with evaluation already in the orchestrating tick, the agent added only the invisibility.

## Verification

- vitest src/test/skills/ — 288 passed (new `pr-followup-arming.test.ts` asserts visibility, naming, drain-before-watch order, watermark-passed-in, all three arming call sites, and zero decision semantics in arm-watcher.md)
- check-skill-deps — no reverse dependencies

https://claude.ai/code/session_01AbQrUE8QdjskUwqbVmj8am


plugin/agents/pr-watcher.md
stan4git, line 1: my question is, without the agent, would the skill still respects the model pin? The very reason to go with agent is the model pin

stan4git: It never did respect it — that's the part the agent detour taught us. Measured under the old cron substrate: 118 tick turns, 0 on haiku; the skill's model pin is inert for dispatched ticks, which was the very motivation for the agent. The ledger now: idle polling runs NO model at all (bash loop — cheaper than haiku, $0); the event wake runs in the orchestrating session, but that turn happens regardless because the dev cycle it dispatches is main-session work — the tick evaluation is one GraphQL call riding a turn that was already paid for. The agent's pin genuinely worked, but after the dumb-poller simplification it was protecting a component that evaluates nothing, at ~40K tokens per quiet cycle plus the invisibility. SKILL.md keeps model: haiku for its own skill-invoked turns (bootstrap, help, reconcile).

stan4git: I am curious about the no cost statement: how would the poll decide if an event needs response then?

stan4git: It doesn't decide — that's the trick. The poll is pure arithmetic: integer comparisons against the watermark file (review id, comment id, unresolved-thread set). It cannot tell whether activity needs a response, only that provider state moved past the floor — which costs zero model tokens. The judgment happens after the wake: the tick re-derives the actionable set from live state with the real semantics (marker rule, resolved/outdated, attempt budgets) and decides reply / dispatch / ignore. So detection is free arithmetic, judgment is one model turn paid only when something actually moved, and the post-cycle watermark advance is what keeps the arithmetic honest — handled activity drops below the floor, so the loop never wakes anyone for it.

stan4git: How does the monitor handles multiple PRs then? showing multiple monitors?

How to name the monitor so that user know which PR it is tracking?

stan4git: One monitor per PR — per-PR isolation was already the rule, and it carries over: N watched PRs show as N running tasks, each reporting and terminating on its own PR's lifecycle independently (auto-track arms one per tracked PR). Naming is the label itself: every monitor is armed as 'PR #<n> — <title>', so the task list reads as a roster of exactly which PRs are being watched, and each event notification carries the same label. Both are in arm-watcher.md; the label format is build-asserted in pr-followup-arming.test.ts.

stan4git: what about blocked on user's decision/call? what is the behavior?

stan4git: Entering a block is a tick decision: it flags last_seen.blocked and emits the one-line owner reminder. While blocked, the monitor stays visible and quiet — the watch keeps standing, costing nothing. The block re-evaluates at the next wake, and here is the honest caveat this question exposes: the monitor watches comment activity and terminal state only, so a block cleared by a push or a CI flip is noticed at the next comment event or arming, not the moment it clears — and the old behavior (a reminder line every minute, which you required in #315) has no heartbeat to ride under the monitor substrate. Two ways to go: (a) accept quiet-while-blocked — zero cost, consistent with the-poller-watches-comments; or (b) the monitor emits a periodic BLOCKED line (say every 10 minutes) while the block flag stands — visible nagging, one wake per reminder. Say which and it ships; SKILL.md's blocked paragraph gets rewritten to match either way.

stan4git: Correction to my earlier 'build-asserted' claim — you were right to push. The test pinned the naming rule in the docs, not on any armed task, and surfaces that show a task's command displayed a bash blob instead of the PR. Fixed: the loop now lives at the slot's watch.sh and the monitor arms with that one-line path, so every surface — label or command — identifies the watch by its PR slot. Live watch re-armed in that form.

stan4git: Deal — shipped. One reminder per block, emitted by the tick that flags it; every later blocked tick is silent while the watch stays visible. Fingerprint auto-resume unchanged: any push, review, or CI/deploy movement clears the block at the next wake, and a re-flag after a resume counts as a new block with its own single reminder. SKILL.md, contract.md Steps 2.5/7, and blocked-tick.md all updated to match.

PR #335, opened 2026-07-23
feat(guardrails): hold the stop until the post-merge next-options offer runs

stan4git:
## Gap

On 2026-07-23 a PR merged mid-session and the session ended with only a status line — no watcher-slot finalization, no teardown, no next-options offer. The post-merge handoff exists as skill guidance but nothing enforced it, so it silently skipped. This productizes the fix as a guardrail for all muggle-works users.

## Design

**Detection (PostToolUse, Bash + Monitor)** — `detectPrTerminal` matches only the verb-phrase output shapes of an actual terminal transition:

- `✓ Merged pull request [owner/repo]#N` (plus the `Squashed and merged` / `Rebased and merged` strategy variants) and `✓ Closed pull request [owner/repo]#N` — both old (bare `#N`) and new (owner/repo-prefixed) gh shapes, scanned across stdout/stderr/output/content since gh writes its success line to stderr
- `TERMINAL pr=N: MERGED|CLOSED` — the pr-followup watch monitor's exit line

It deliberately never matches bare `"state":"MERGED"` metadata (every `gh pr view --json` carries it) or git's `Merge pull request #N` commit subject. On detection the PR number lands in `terminalPending` (deduped against pending and already-handled) and a PostToolUse envelope directs the handoff: finalize the watcher slot, tear down per `autoCleanup`, then offer next options via AskUserQuestion.

**Satisfaction** — a PostToolUse firing on AskUserQuestion (`offer-ran`) is the only exit: it moves pending PRs to `terminalHandled` and resets the block counter. The handled list keeps a replayed terminal line (log tail, monitor notification replay) from re-arming a gate the offer already cleared.

**Stop gate** — `terminal-gate` blocks the turn end while `terminalPending` is non-empty, naming the pending PR(s); full instruction on the first block, one-line reminders after; unconditional release after 3 blocks (`MAX_PR_TERMINAL_BLOCKS`).

## No counter reset

The e2e gate's counter resets on every passing unit test, which turned its 3-block cap into ~20 nags in one long session. Here nothing resets `terminalBlockCount` except the offer itself — not a unit test, not a tick, not another Bash call. `prTerminal.test.ts` pins this with a regression test that runs `applyRecordedRun({unitTestPassed: true})` against an armed gate and asserts the counter is untouched.

## Wiring

- `hooks.json`: `guardrail-pr-terminal.sh` joins the Bash PostToolUse fan-out (keyword pre-filter keeps Node off the hot path) and a `Monitor` matcher; `guardrail-offer-ran.sh` on `AskUserQuestion`; `guardrail-terminal-gate.sh` on `Stop` — both state-file pre-filtered so turns with no terminal PR never spawn Node.
- Bundle `plugin/scripts/guardrails.mjs` rebuilt via `npx tsup` (never hand-edited); rebuild is deterministic — identical sha256 across two builds.

## Test evidence

- `npx vitest run src/test/guardrails` — 9 files, 97 passed, 8 skipped (win32-skipped bash pre-filter suite), exit 0
- `npx eslint src/guardrails/ src/test/guardrails/` — exit 0; `npx tsc --noEmit` — exit 0; `node scripts/check-skill-deps.mjs` — OK, 311 links, no reverse deps
- End-to-end through the real wrappers + rebuilt bundle (Git Bash, temp HOME): detect → nudge, dedupe repeat → `{}`, Stop blocks 1/2/3 with escalating-terse reasons → 4th releases `{}`; offer clears an armed gate and a replayed `TERMINAL pr=N` line afterwards stays quiet

https://claude.ai/code/session_01AbQrUE8QdjskUwqbVmj8am

### DIFF
--- a/src/guardrails/prTerminal.ts
+++ b/src/guardrails/prTerminal.ts
@@ -1,0 +1,117 @@
+import type { GuardrailState, HookInput } from "./types.js";
+
+export enum PrTerminalVerdict {
+  Merged = "merged",
+  Closed = "closed",
+}
+
+export interface PrTerminalEvent {
+  prNumber: number;
+  verdict: PrTerminalVerdict;
+}
+
+// gh success lines: `✓ Merged pull request #12 (title)` — also the
+// "Squashed and merged" / "Rebased and merged" strategy variants — and
+// `✓ Closed pull request owner/repo#12 (title)`. Newer gh prefixes the number
+// with owner/repo, older prints a bare `#12`; the optional [\w./-]* run accepts
+// both. Anchoring on the full verb phrase (never a bare MERGED/CLOSED token) is
+// what keeps `"state":"MERGED"` out: every `gh pr view --json`/API fetch
+// carries that token, and matching it would arm the gate on a routine status
+// query. It also excludes git's own merge-commit subject ("Merge pull
+// request #N"), which differs by verb tense.
+const GH_MERGED_LINE = /\b(?:Merged|Squashed and merged|Rebased and merged) pull request [\w./-]*#(\d+)/;
+const GH_CLOSED_LINE = /\bClosed pull request [\w./-]*#(\d+)/;
+// The pr-followup watch monitor's exit line (e.g. `TERMINAL pr=331: MERGED`),
+// which also surfaces when a monitor event notification is replayed through a
+// tool result.
+const MONITOR_TERMINAL_LINE = /\bTERMINAL pr=(\d+): (MERGED|CLOSED)\b/;
+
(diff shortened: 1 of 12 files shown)


src/guardrails/prTerminal.ts
stan4git, line 1: better organize your code please

stan4git: Reorganized per repo conventions: the four PrTerminal types/enums now live in types.ts, the detection regexes and MAX_PR_TERMINAL_BLOCKS in a new constants.ts (rationale comments kept with the regexes), leaving prTerminal.ts logic-only — aa45e90.

PR #343, opened 2026-07-23
fix(pr-followup): keep an outdated thread actionable while its question is unanswered

stan4git:
## Gap

The watcher's actionable-thread trigger required `isOutdated == false`. A thread whose anchored line moved — a rebase, or any later push to the branch — was dropped, **even when its newest comment was an unanswered human question**. No tick ever surfaced it, so the question sat with no response.

This bit for real: a `why isn't this returning X` review comment landed ~7 min before a force-push that carried a rebuild. The push moved the line, GitHub marked the thread outdated, and the next drain tick reported "0 actionable" in good faith — leaving the question unanswered until it was noticed by hand.

## Fix

Outdatedness is a poor proxy for "already handled" — it means the line moved, not that the concern was addressed. The **marker rule already prevents re-dispatch** once the loop replies (the thread's newest comment becomes the loop's own, marker-carrying, comment), so resolution + marker are the correct and sufficient gates. Whether the moved code mooted the concern is a judgment `/muggle-do` makes and *replies to* — not a reason for the watcher to stay silent.

- `contract.md` Step 3(a): drop `isOutdated == false` from the actionable trigger; a thread is actionable when `isResolved == false` and its newest comment lacks the loop marker.
- `_shared/vcs/github/unresolved-threads.md`: same — the dispatch trigger no longer requires `isOutdated == false` (the resolve-reminder stage already ignored it).
- Update the two `pr-followup-tracking-flow` assertions to the new rule.

Echo is still intrinsic: once the loop replies (marker), the thread is non-actionable regardless of outdatedness, so no re-dispatch loop.

## Testing

`vitest run src/test/skills` — 290 passed. No residual references to the old gate.


plugin/skills/_shared/vcs/github/unresolved-threads.md
stan4git, line 1: let's add eval cases for this case

stan4git: Added in `66650ca` — two tick-decision evals in `muggle-pr-followup/evals/evals.json` covering both directions of the gate:

- **id 9** `outdated-thread-with-unanswered-question-still-actionable` — thread `isOutdated=true`, newest comment an unmarked human question → plan must mark it actionable and dispatch address-reviews, and must NOT skip on outdatedness.
- **id 10** `outdated-thread-loop-already-replied-stays-quiet` — thread `isOutdated=true`, newest comment carries the `muggle-do:bot` marker → plan must idle (the marker rule holds regardless of outdatedness; no echo re-dispatch).

Together they pin that the gate is resolution + marker, never outdatedness. Full skills suite green (290).

PR #345, opened 2026-07-23
fix(pr-followup): stop watcher and MCP process leaks that freeze Claude Code on Windows

stan4git:
## Problem

On Windows, muggle-works leaks two kinds of child process that outlive the Claude session that launched them, and they accumulate across sessions until the machine is storming with `gh` / `bash` / `node` processes — freezing Claude Code. Observed live on one machine: **18 orphaned `watch.sh` loops** (6 logical watchers from 4+ dead sessions, each waking every 60s to fire ~4 `gh` calls) and **45 zombie MCP `node` servers** (some 6 days old), plus **646** never-GC'd guardrails JSON files and **257** stale watcher slots.

Root cause is one class of bug: a child that doesn't die when its Claude session does.

## Fixes

**1. Watcher lifecycle + dedup** — the per-slot watch loop was a bare `while true` with no orphan or dedup guard, and the "arm only when none alive" check inspected the in-session monitor, not the OS process.
- New shipped, unit-tested guard library `plugin/scripts/pr-watch-guards.sh`. The loop claims the slot via `watch.pid` and each iteration exits if **superseded** by a newer arm or past a **6h lifetime cap** — never unbounded.
- `arm-watcher.md` now does real cross-process dedup (skip arming when a live `watch.pid` owns the slot); `reconcile.md` no longer arms a recovery cron on top of a live monitor. `state-schemas.md` documents `watch.pid`.

**2. MCP zombie** — `stdin` end/close (already shipped) never fires on *abrupt* Windows parent death (kill/crash/sleep), so the stdio server leaked. `src/server/stdio-server.ts` adds an active **PPID-liveness poll** that exits once the original parent is gone.

**3. State GC** — nothing pruned the per-session guardrails files or finalized watcher slots. New TTL-gated `plugin/scripts/gc-state.sh` (SessionStart) prunes `guardrails/*.json` >14d and finalized slots >30d, never touching an open slot or the live session.

## Verification

- `pnpm test` (changed areas): **23/23 pass** — `stdio-server` (PPID poll: exits on parent-gone, stays up while alive, no-op when ppid≤1), `pr-watch-guards` (supersede / lifetime / pid-alive), `gc-state` (prunes stale, keeps fresh + open slots + honors the TTL gate).
- `tsc --noEmit` clean, `eslint` clean on changed files, `check-skill-deps` OK (no reverse deps).

## Operational note

The already-running orphans on the affected machine were killed manually as immediate relief. Orphaned per-minute recovery `/loop` crons from dead sessions are invisible to `CronList` and only clear on a **Claude Code restart** — recommended after merge so watchers consolidate to one live set under the fixed logic.


plugin/hooks/README.md
```diff
@@ -50,3 +50,7 @@ Each guardrail is a thin bash wrapper in `../scripts/` registered in `hooks.json
...
 
 Cost: zero open slots — the common case — is one directory scan, no `node` spawn. With open slots the hook adds one short-lived `node` liveness check (tens of ms). The daemon is a single idle process waking every 5 minutes for local file stats; it makes provider API calls only for a slot whose liveness beacons are stale (2–4 per dead slot per scan, zero while watchers are live) and spends model tokens only when a recovery tick actually spawns.
+
+## Session-start state GC
+
+`SessionStart` (`scripts/gc-state.sh`) — prunes ephemeral state that nothing else garbage-collects, so it doesn't grow without bound (the per-session guardrails files and finalized watcher slots otherwise accumulate one-per-session forever). Deletes `~/.muggle-ai/guardrails/*.json` older than 14 days (each is read only by its own session's hooks) and finalized watcher slots — `result.md` present — older than 30 days (their `followup.log` is forensic-only). TTL-gated to once per day via a `~/.cache/muggle/state-gc-checked` marker; silent and best-effort, never blocks session start. Never touches an open slot (no `result.md`) or the current session's own state.
```
stan4git: I do have changes or sessions that lasts more than a few months. Please use a better design

stan4git: Good catch — fixed in d35f1b4.

Reworked `gc-state` so collection is keyed on **inactivity, not creation age**. Every session start now refreshes its own `guardrails/<session>.json` (on top of `guardrails.mjs` rewriting it on each guarded tool call), so a session that runs for months keeps a live mtime and is never swept — only a session gone quiet for the *whole* window (i.e. ended) is collected. Open watcher slots (no `result.md`) were already never pruned, so a PR watched for months is safe too.

Both windows are now env-overridable (`MUGGLE_GUARDRAILS_TTL_DAYS` / `MUGGLE_SLOT_TTL_DAYS`), and there's a test asserting a 60-day-old file for the *running* session survives while an equally-old dead one is collected.

PR #344, opened 2026-07-23
fix(skills): never push unsigned commits

stan4git:
## Goal

The muggle skills must never push unsigned commits. The do/pr-followup flows committed and pushed with no signing check — on a machine without a working signing setup, every cycle would push unsigned.

## Acceptance Criteria

- A shared recipe defines the never-push-unsigned rule: `%G?` preflight, server-signed commits via GraphQL `createCommitOnBranch`, rebase replay for the force-push path, GitLab escalation (no server-side signing analogue).
- Every push path routes through the gate: `push-to-branch.md` (address-reviews push), `open-prs/forward.md` (first push), `build.md` (commit), `resolve-conflicts.md` (force-push).
- Disabling signing (`--no-gpg-sign`, `commit.gpgsign=false`) is never an option.
- Static lint test locks the wiring.

## Changes

- New `plugin/skills/_shared/vcs/github/signed-commits.md` — the recipe.
- `push-to-branch.md` — signing gate before the push commands.
- `github.md` index — updated push-to-branch row + new signed-commits row.
- `do/build.md` — commit step preflights signing, routes unsigned machines to the remote path.
- `do/open-prs/forward.md` — Stage 7 push goes through the gate.
- `do/resolve-conflicts.md` — force-push blocked when unsigned; replay path + guardrail.
- New `src/test/skills/signed-commits-recipe.test.ts` — static wiring lint.

## Validation

unit-only — static lint test (`npx vitest run src/test/skills`: 298 passed; `node scripts/check-skill-deps.mjs`: OK; eslint clean). E2E skipped: no web surface. Dogfood: this branch's commit was itself created server-signed via `createCommitOnBranch` per the new recipe (no local signing on this machine).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018suY8uimMSxsVnsRFtSVw2

PR #341, opened 2026-07-23
[UNIT-ONLY] pr-followup: out-of-session watchdog recovers dead watchers

stan4git:
## Goal

muggle-pr-followup watchers survive orchestrating-session death (usage-limit hit, session end) — open PRs regain a poller automatically after the limit resets, with zero user action.

## Acceptance Criteria

- Detached singleton watchdog daemon outside any Claude session scans open slots on an interval (default 300s) and exits when none remain
- Watch loops touch `<slot>/watch-heartbeat` each iteration; liveness = heartbeat or newest `followup.log` line within 15 min — a quiet-but-alive monitor is never treated as dead
- Dead slots are polled with plain `gh` calls (no AI): state, actionable threads (loop-marker rule), body-only reviews past the watermark, behind/conflict, CI rollup
- Recovery entry is the existing tick mode, spawned headless (`claude -p "/muggle:muggle-pr-followup <slug> <n>"`); no new skill mode
- Spawn discipline: terminal spawns immediately; other signals need two consecutive sightings; an unchanged signature never respawns; a silent spawn (usage limit active) retries on later scans — recovery lands at limit reset; max 3 spawns/scan
- A slot mid-`/muggle-do`-cycle (dispatch line newer than last outcome line, < 90 min old) is deferred
- Ensured from a SessionStart hook (`ensure-pr-watchdog.sh`) and from arm-watcher step 4
- Reconcile Step 3.6 liveness switched to the heartbeat-or-log beacon rule

## Changes

- `src/watchdog/` — decision logic (liveness, followup.log inference, signature/actionability, spawn decision) + daemon I/O shell; bundled by tsup to `plugin/scripts/pr-followup-watchdog.mjs` (committed, like guardrails.mjs)
- `plugin/scripts/ensure-pr-watchdog.sh` + SessionStart registration in `plugin/hooks/hooks.json`
- Skill docs: `SKILL.md` (Session death), `arm-watcher.md` (heartbeat beacon + arm-time ensure), `reconcile.md` (§ Out-of-session watchdog + Step 3.6 beacon liveness), `state-schemas.md` (`watch-heartbeat`, `watchdog.json`), hooks `README.md`
- Tests: `src/test/watchdog/*` (28 cases over the decision modules) + `src/test/skills/pr-followup-watchdog-hook.test.ts` (static wiring lint)

## Validation

[...]

PR #347, opened 2026-07-24
[UNIT-ONLY] pr-followup: dead watcher masked by non-tick log lines

stan4git:
## Goal
A dead poller must never read as alive, and the main session must never poll unwoken. Reconcile and the watchdog counted any parseable `followup.log` line as a liveness beacon, so arming announcements (`armed …`/`re-armed …`), cycle notes, and error lines — written by sessions that can die the next instant — suppressed re-arm for the whole 15-minute window (observed live on #344). Separately, a manual `<slug> <n>` invocation routed straight to the tick and ran a main-session provider poll nothing prompted.

## Changes
- `src/watchdog/followupLog.ts`: `newestFollowupLogTimestampMs` → `newestTickLineTimestampMs` — only `<ISO> tick …` / `<ISO> stale-tick …` lines count (second-token anchored, so a hypothetical "watchdog spawned recovery tick" line stays inert).
- `liveness.ts` / `decide.ts` / `types.ts` / `cli.ts`: beacon and tick-ran-after-spawn both use the tick-only timestamp; bundled `plugin/scripts/pr-followup-watchdog.mjs` rebuilt.
- `reconcile.md` Step 3.6 + watchdog paragraph: tick-line beacon documented (merged over #349's watch.pid-lease-first structure) — logging is not polling.
- `arm-watcher.md` step 5: the poll/wake split stated explicitly — polling never enters the session (cadence = detached monitor loop + out-of-session watchdog); the session receives only the monitor's single event line, and that wake turn runs the tick + cycle inline in the owning session, never a subagent (owner direction, 2026-07-24/28).
- `SKILL.md` routing: **live-watcher gate** — `<slug> <n>` with a live `watch.pid` and no `--wake` routes to **watch-status** (on-disk report, zero provider calls); event wakes pass `--wake=<event>`; recovery fires reach the tick because a recoverable watcher is dead by definition; stale crons absorb as a status line.

## Validation
unit-only — full vitest 899 passed (69 files) incl. incident-shaped beacon cases (fresh non-tick lines ⇒ dead; tick/stale-tick ⇒ alive), doc-locks on reconcile wording, the inline-wake rule, and the live-watcher gate; `verify:skill-deps` green; all commits server-signed via createCommitOnBranch.

https://claude.ai/code/session_01Sz9jfMQWttuyhCR5x3qJQ7


plugin/skills/muggle-pr-followup/reconcile.md
```diff
@@ -16,7 +16,7 @@ Recover-don't-seed holds on every trigger: a session-start run still never seeds
...
+Every trigger above needs a live session, so a watcher killed by a session's death — a usage-limit hit ends the monitor stream mid-watch — stays dead until a human starts the next session. The watchdog daemon (`plugin/scripts/pr-followup-watchdog.mjs`, ensured by the `ensure-pr-watchdog.sh` session-start hook and by [`arm-watcher.md`](arm-watcher.md) step 4) closes that gap from outside any session: a detached singleton that scans open slots on an interval, treats a slot as live on the same heartbeat-or-tick-line beacon as Step 3.6, polls dead slots with plain provider-CLI calls (`gh` / `glab`, per the slot's `prs.json` `provider`), and spawns a **headless recovery tick** (`claude -p "/muggle:muggle-pr-followup <slug> <n>"`) when a dead slot has a terminal PR or a signature-confirmed signal. A spawn that dies silently (usage limit still active) is retried on later scans — recovery lands on its own once the limit resets. The tick, not the watchdog, holds all semantics; the watchdog only decides *that* a tick is owed. It defers to live watchers, holds off while a dispatched cycle owns the PR, never re-spawns an unchanged signature, and exits when no open slot remains.
```
stan4git: I don't like that you are introducing race condition here. why you have a watchdog to run a detached process outside of the session???

I expect the prfollow up goes with the session activiness. session down, watchers down. session up, watchers run (if PR is out)

stan4git: Agreed — and this session proved your point: the watchdog spent the last few hours spawn-retrying this very PR every ~10 min (headless `claude -p` ticks that never landed on Windows), and yes, there's a real race — the beacon read and the spawn decision aren't atomic, so a stale-beacon scan can spawn a headless tick just as a live session arms its monitor, and both dispatch a cycle → double replies/pushes on one PR.

Why it existed: #341 added it to cover the one gap session-scoped watching can't — a usage-limit kill or closed laptop while reviews land, so nobody's watching until you open the next session. That's the only thing removing it costs.

Your model — **session down → watchers down; session up → reconcile re-arms every open PR** — is already what the in-session machinery does; the watchdog is a bolt-on. Removing it:
- delete `src/watchdog/*`, the built `plugin/scripts/pr-followup-watchdog.mjs`, and the `ensure-pr-watchdog.sh` hook + its tsup entry;
- drop the "ensure watchdog" step from `arm-watcher.md` and the whole "Session death"/"Out-of-session watchdog" sections here and in `SKILL.md`;
- keep the existing session-start `reconcile-stale-watchers.sh` hook — on session start it re-arms every open slot whose monitor is dead (which, session-scoped, is all of them), giving exactly "session up → watchers run."

One tradeoff to confirm before I cut it: a review that lands while **no** session is running waits until your next session start. If that's acceptable (your message reads like it is), I'll rip the watchdog out on this PR.

stan4git: "a review that lands while no session is running waits until your next session start." yes this is by design. I don't want to address comments without the session's context

stan4git: Done — watchdog removed in 854729d. `session down → watchers down; session up → reconcile re-arms every open PR`, exactly your model.

Cut:
- `src/watchdog/*` + all its tests, the bundled `pr-followup-watchdog.mjs`, the `ensure-pr-watchdog.sh` SessionStart hook and its `tsup` entry;
- every watchdog reference in `arm-watcher.md` / `reconcile.md` / `SKILL.md` / `contract.md` / `state-schemas.md` / `stop.md` and the evals.

Kept: the session-start `reconcile-stale-watchers.sh` nudge — on session start it re-arms every open slot whose monitor is dead, which is the whole recovery path now. No detached process, nothing polls or replies without a session's context.

Locked it so it can't creep back: a doc-lock test fails if any skill doc names a watchdog or a headless `claude -p` tick again. Full suite 833 green (the 66-test drop is the deleted watchdog suites), typecheck + skill-deps clean, all commits server-signed.

PR #350, opened 2026-07-28
feat(guard-run): kernel-capped process-tree launcher — spawn-time storm guard

stan4git:
## Why

Incident 2026-07-27: a killed agent's orphaned CRA jest watch run re-executed its bootstrap into ~12,500 node processes on Windows, starving the machine and killing the user's dev servers. A reactive sweeper is always too late — the guarantee has to be enforced at spawn. This PR gives works a kernel-enforced spawn-time cap plus the repo's own watch-mode gates.

## What

**`guard-run` — portable capped launcher** (`plugin/scripts/guard-run.mjs`, built from `src/guard-run/`)

`guard-run [--limit N] [--service] -- <command...>` runs the command with its entire process tree kernel-capped (default 64): the process past the limit fails at creation, nothing to clean up. Default mode also tethers the tree to the launcher — killing guard-run reaps every descendant. `--service` keeps the cap but lets the tree outlive the launcher (dev servers).

Backends, dispatched per platform:

| Platform | Mechanism | Tree death with launcher |
| :--- | :--- | :--- |
| win32 | Job Object shim (`guard-run-job-object.ps1`, Add-Type P/Invoke): child created `CREATE_SUSPENDED`, assigned, resumed — jailed from its first instruction; `JOB_OBJECT_LIMIT_ACTIVE_PROCESS` + `KILL_ON_JOB_CLOSE` + launcher-death watch (`WaitForMultipleObjects` on child + launcher, `TerminateJobObject` if launcher dies first) | kernel-guaranteed |
| linux | `systemd-run --user --scope -p TasksMax=N` (cgroup v2 `pids.max`, excess forks EAGAIN); fallback `prlimit --nproc` + new process group | scope stopped / group killed on catchable exit; cap itself survives even SIGKILL |
| darwin | `ulimit -u` headroom in the child shell + new process group | group killed on catchable exit |

rlimit backends cap per-user (current count + limit headroom) — documented as looser than the cgroup/Job Object per-tree caps. If no backend can enforce a cap, guard-run refuses to run the command — never silently unguarded.

**Repo's own watch gate** — `vitest.config.ts`: bare `vitest` now runs one-shot; watch is opt-in via `MUGGLE_VITEST_WATCH=1`. (Only vitest config in the repo; no jest configs.)

**Skill guardrails** — `do/unit-tests.md`: suites run one-shot (`CI=true` / `--watchAll=false` / `vitest run`), never watch, launched through guard-run. `muggle-test-prepare/steps/start-services.md`: services launch through `guard-run --service`. Uniform `guard-run` reference, no per-OS recipes.

## Verification

- 934 tests green (`npx vitest run`), 41 new dispatcher tests (parsing, backend selection per injected platform, headroom, per-backend command construction — pure, no real spawning). `tsc --noEmit` and eslint clean; `check-skill-deps` and `verify-plugin-marketplace` pass.
- Live Windows smoke: exit code 7 forwarded end-to-end; fork script spawning 100 children under `--limit 10` → 9 spawned + parent = 10, **91 blocked at creation**, 0 survivors after run; SIGKILL of the launcher mid-run → whole tree AND shim gone in <4s; `--service` tree survived launcher SIGKILL, still capped.
- POSIX smoke procedure documented as a `describe.skip` in `buildLaunchPlan.test.ts` for a linux/darwin host.

https://claude.ai/code/session_016onpcHAZJxbCCYh4riDbo8


plugin/scripts/guard-run-job-object.ps1
stan4git, line 1: how is this going to be os agnostic?

plugin/skills/muggle-test-prepare/steps/start-services.md
stan4git, line 1: how is this os agnostic??

plugin/scripts/guard-run-job-object.ps1
stan4git: This file is never what callers touch — the entry point is `guard-run.mjs`, plain node, identical invocation on every OS. It probes the platform and dispatches to one backend: this Job Object shim on win32, `systemd-run --scope TasksMax` (or `prlimit` fallback) on linux, `ulimit -u` + process group on darwin. PowerShell exists only inside the win32 backend, exactly as `systemd-run` exists only inside the linux one. Header now states this up front: 43dd5ac.

plugin/skills/muggle-test-prepare/steps/start-services.md
stan4git: The instruction itself is the OS-neutral part: wrap the start command as `node <plugin-root>/scripts/guard-run.mjs --service -- <command>` — same wrapper everywhere, because guard-run.mjs picks the cap mechanism per OS internally. The bash block is just that instruction rendered for the Bash tool (which this step already used pre-PR, and which runs as Git Bash on Windows too). Rephrased in 43dd5ac so the wrap is defined as the instruction, with the recipe as its rendering.

test/guard-run/buildLaunchPlan.test.ts
stan4git, line 1: why all these test files under src/*???

stan4git: These follow the repo's existing test convention — tests live under `src/test/<area>/` (e.g. `src/test/cli/*.test.ts`, `packages/mcps/src/test/`), so the guard-run suites sit at `src/test/guard-run/` to match. Happy to relocate if you'd prefer a top-level `test/` root, but that would be a repo-wide move rather than something specific to this PR.

stan4git: No, this is not the right convention. "src" should be the save level folder as "test" (with the same relative path under) to match code file and test file.


Fix everything that caused this error

stan4git: Done — moved the four guard-run suites so `test/` sits beside `src/` at the repo root, each file at the same relative subpath:

- `src/test/guard-run/buildLaunchPlan.test.ts` → `test/guard-run/buildLaunchPlan.test.ts`
- `src/test/guard-run/parseArguments.test.ts` → `test/guard-run/parseArguments.test.ts`
- `src/test/guard-run/selectBackend.test.ts` → `test/guard-run/selectBackend.test.ts`
- `src/test/guard-run/headroom.test.ts` → `test/guard-run/headroom.test.ts`

(via `git mv`, so history follows.) Imports go up one more level accordingly: `../../guard-run/X.js` → `../../src/guard-run/X.js`.

vitest `include` now lists Vitest's default glob **plus** `test/**/*.test.ts` — I kept the default first so `packages/**` and `internal/**` stay discovered (verified: discovery set is byte-identical to before, so nothing else moved). Coverage still excludes `src/guard-run/cli.ts` and the guard-run source stays instrumented from `src/`.

Scope kept to this PR's own tests — the repo-wide `src/test/**` layout is untouched. `npx vitest run` green (guard-run 41 passed / 1 skipped at the new path), `verify:skill-deps` + `verify:plugin` pass. Pushed 87a2728.

PR #348, opened 2026-07-27
[UNIT-ONLY] Capture memory-exhaustion debug context (fleet + host)

stan4git:
## Goal

When the next memory-exhaustion event happens on the fleet host, enough context is captured automatically to name the consuming processes, their parentage, and the growth timeline — the three things the 2026-07-26 freeze post-mortem could not recover (Windows' Resource-Exhaustion-Detector OOM'd before logging any consumer, the daemon reap abandoned children namelessly, and watch-loop stderr carried no timestamps).

## Acceptance Criteria

- Watchdog writes an orphan ledger (`orphan-ledger-<ts>.json`) into a slot when it finds the watcher dead: suspected orphaned heavy processes (parent gone from the process table), max once per 6h window, nothing written when no suspects.
- `arm-watcher.md` requires UTC-timestamped lines in `watch-stderr.log` (no bare `2>>` redirect).
- muggle-do stages append `workload` ledger lines to `followup.log` for heavy spawns (test suites, typechecks, dev servers); format documented in `state-schemas.md`.
- `ops/memwatch`: per-minute sampler + 85%-commit tripwire (Scheduled Task, unelevated), perfmon circular flight recorder + boot autostart, and 4688 process-creation audit with command lines (elevated installer, GUID subcategory for locale safety).

## Changes

- `src/watchdog/orphanLedger.ts` (+ vitest suite, wired into `cli.ts` scanOnce; bundle regenerated): process-table parsers (Win32/POSIX), orphan-suspect selection, dedup-windowed ledger write.
- `plugin/skills/muggle-pr-followup/{arm-watcher,state-schemas}.md`: timestamped stderr requirement; `watch-stderr.log`, `orphan-ledger`, and `workload`-line schemas.
- `plugin/skills/do/{build,unit-tests,e2e-acceptance}.md`: workload-ledger instruction at each heavy-spawn point.
- `ops/memwatch/`: `memwatch.ps1` sampler/tripwire, `install-memwatch.ps1` (installs to `~\.muggle-ai\memwatch\bin`, survives checkout deletion), `install-elevated.ps1` (logman DCS + audit policy, writes machine-readable result), README with post-mortem reading guide.

## Validation

[...]

PR #352, opened 2026-07-29
feat(ops): host-level memwatch memory sampler + commit tripwire

stan4git:
## What

Host-level memory-diagnostics tooling under `ops/memwatch/` — the standalone, watchdog-free half of #348, so we can actually debug the memory-exhaustion freezes on the fleet host.

- **`memwatch.ps1`** — per-minute sampler: `% Committed Bytes In Use` + committed/free GB + top-15 processes by private bytes → daily JSONL. At ≥85% commit it writes a full process snapshot (pid/ppid/parentAlive/priv/ws/cmd) and pops a toast, re-arming below 75%. 14-day sample retention, 90-day snapshot retention.
- **`install-memwatch.ps1`** — registers the unelevated `MuggleMemwatch` scheduled task and copies the script to `~/.muggle-ai/memwatch/bin` so it survives deleting the checkout/worktree.
- **`install-elevated.ps1`** — optional: perfmon circular flight recorder + boot autostart, and a 4688 process-creation audit with command lines.
- **`README.md`** — post-mortem reading guide.

## Why now

The 2026-07-27 22:45 host freeze hit 99% commit and hard-reset before Windows' RADAR could log a consumer; the sampler had been disabled, so the 40-minute runaway went uncaptured. This lands the tooling in-repo so it's tracked and improvable.

## Not shipped

`ops/` is outside the npm `files` whitelist (`dist`, `plugin`, `bin/muggle.js`, `scripts/postinstall.mjs`), so none of this reaches `@muggleai/works` end users. Internal host tooling only.

## Supersedes #348

#348's watchdog half (`src/watchdog`, orphan ledger) was removed upstream in `854729d` (session-scoped watching). This PR carries only the still-relevant host tooling; #348 will be closed.

### DIFF
--- a/ops/memwatch/toggle-memwatch.ps1
+++ b/ops/memwatch/toggle-memwatch.ps1
@@ -1,0 +1,40 @@
+# On/off switch for the MuggleMemwatch sampler task. Wraps Enable/Disable so the
+# tracker can be paused mid-investigation without unregistering it — samples and
+# trip-state survive, and -On resumes on the next minute. Unprivileged (the task
+# is registered per-user); PowerShell 5.1-compatible. The elevated flight-recorder
+# and audit pieces are not covered here — see README "Uninstall" for those.
+[CmdletBinding(DefaultParameterSetName = 'Status')]
+param(
+    [Parameter(ParameterSetName = 'On', Mandatory)][switch]$On,
+    [Parameter(ParameterSetName = 'Off', Mandatory)][switch]$Off,
+    [string]$TaskName = 'MuggleMemwatch'
+)
+
+$ErrorActionPreference = 'Stop'
+
+$task = Get-ScheduledTask -TaskName $TaskName -ErrorAction SilentlyContinue
+if ($null -eq $task) {
+    if ($On) {
+        Write-Warning "$TaskName is not installed — run install-memwatch.ps1 first to register it."
+        exit 1
+    }
+    Write-Output "$TaskName is not installed (nothing to toggle)."
+    exit 0
+}
+
+switch ($PSCmdlet.ParameterSetName) {
+    'On' {
+        Enable-ScheduledTask -TaskName $TaskName | Out-Null
+        Write-Output "$TaskName enabled — sampling resumes within a minute."
+    }
+    'Off' {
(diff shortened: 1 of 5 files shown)


stan4git:
Added `toggle-memwatch.ps1` — a reversible on/off switch for the sampler (`-Off` / `-On` / no-arg status), wrapping Enable/Disable-ScheduledTask so it pauses without unregistering or losing samples/trip-state. README gains a **Pause / resume** section and Uninstall now points to it for the non-destructive path. Parse-checked and Status-run verified; commit is signed (`2b41b47`).

ops/memwatch/toggle-memwatch.ps1
stan4git, line 1: Is it possible to bring this in to CI to gate keep potential problems while running evals/tests

stan4git: Good idea — possible, but not by running this script as-is. `memwatch.ps1` is Windows host telemetry (perf counters + WMI + toast), so it won't run on the Linux CI jobs, and a container's commit-% isn't the same signal as a dev box.

For a CI gate I'd wrap the evals/tests step in a portable **peak-memory** check that fails the job over a budget: on Linux runners `/usr/bin/time -v` "Maximum resident set size" (or a `ulimit -v` / cgroup ceiling so an exhaustion run OOM-fails deterministically), plus a small Node sampler if we want the same gate on the `windows-latest` matrix leg. Tradeoff: tight thresholds add flakiness, so I'd land it **non-blocking** (report only) first, then flip to blocking once we have a few baselines.

Want me to spec that as a separate follow-up (not this toolkit PR) and wire a `mem-gate` step into the evals/tests workflow?


stan4git: yes as a follow up

stan4git: Done — opened #355 as the follow-up. It adds a portable peak-RSS gate (`scripts/mem-gate.sh`, GNU `/usr/bin/time -v`) wired around `pnpm test` in the `lint-and-test` job, **report-only** for now; a budget can flip it to blocking once we've seen a few baselines. Kept entirely separate from this toolkit PR, as you asked.

PR #353, opened 2026-07-31
fix(pr-followup): advance watermark to the handled wave's snapshot, never a post-reply live-max

stan4git:
## Problem

The `muggle-pr-followup` watcher's watermark-advance could silently **swallow a reviewer comment** (observed live).

The watermark tracks the highest handled review/comment id. `arm-watcher.md`'s "After a cycle" guidance told the operator to advance the watermark from a **fresh live-max fetched *after* replies are posted**. A reviewer comment that arrives during the handling window — between reading the wave and posting the replies — then lands *below* that fresh max and is marked seen without ever being read. It happened: a "show the whole ER design" comment was buried and only surfaced when a human pointed it out.

## Fix

Two invariants, applied wherever the watermark is set or advanced:

1. **Advance to the handled wave's snapshot, not a post-reply live fetch.** Capture the max review-id and comment-id **at the start of handling** (the ids the cycle actually read), and advance to exactly those. Anything arriving after the snapshot stays above the watermark and surfaces on the next tick.
2. **Scan before advancing (hard invariant).** Before advancing, confirm every unresolved review thread in the handled wave was addressed (reply posted or explicit deferral). If any is unaddressed, don't advance past it — leave its comment id above the `COM` floor and its thread id out of the `THREADS` set so the next tick re-surfaces it.

The same snapshot-at-read rule is applied to the arm-time seed and the Step 7.5 re-arm seed (seed to the drain/tick's own observed max, never a fresh later fetch).

## Files

- `plugin/skills/muggle-pr-followup/arm-watcher.md` — Seed step and the "After a cycle" advance (adds the scan-before-advance invariant).
- `plugin/skills/muggle-pr-followup/contract.md` — Step 7.5 re-arm seed.
- `plugin/skills/muggle-pr-followup/state-schemas.md` — `watch-watermark.env` description.
- `plugin/skills/muggle-pr-followup/evals/evals.json` — one assertion updated to the corrected rule.

## Verification

- `npm run verify:skill-deps` — OK (335 cross-skill links, no reverse dependencies)
- `npm run verify:plugin` — passed
- `npx vitest run pr-followup` — 9 files, 144 tests passed

https://claude.ai/code/session_016onpcHAZJxbCCYh4riDbo8


plugin/skills/muggle-pr-followup/arm-watcher.md
```diff
@@ -3,7 +3,7 @@
...
... — every fetch in this sequence uses that provider's recipes. Write the slot's watch watermark ([`state-schemas.md`](state-schemas.md#watch-watermarkenv)) to the ids the **drain itself read** — the max review-id and comment-id observed at the drain's own fetch (Step 1), snapshotted at that read. Never let the loop capture its own baseline — the arming session writes it; and **never** from a fresh fetch taken after the drain, which would include a comment that arrived after the drain read the wave and mark it seen unread. Seeded to the drain's floor, anything landing after that read stays above the watermark and the monitor's first iteration surfaces it. Seed the CI floor (`CIRED`) from the same drain read: set it to the head SHA when the checks have **already settled red** at that read (no check pending, one or more in the `fail` bucket per the provider's rollup recipe — [`../_shared/vcs/github/pr-checks.md`](../_shared/vcs/github/pr-checks.md) / [`../_shared/vcs/gitlab/mr-pipeline.md`](../_shared/vcs/gitlab/mr-pipeline.md)) — that red is what the drain just handled — and empty otherwise, so an escalated red head the drain already saw does not re-fire on the loop's first iteration.
```
stan4git: can you avoid referencing gitlab+github instructions at the same time? make a shared instruction and reference them in the shared instruction

stan4git: Done in d459c7a. Added two tool-agnostic dispatchers under `plugin/skills/_shared/vcs/common/` that own the github-vs-gitlab branch, so each site links one shared instruction instead of naming both recipes:

- **`common/ci-rollup.md`** — the CI poll; dispatches to `github/pr-checks.md` / `gitlab/mr-pipeline.md` and folds into red/pending/green. Now referenced by `arm-watcher.md` (watermark seed + monitor-wake) and `contract.md` Step 6.
- **`common/branch-standing.md`** — conflict + out-of-date (behind-by) detection; dispatches to `github/pr-metadata.md` / `gitlab/mr-metadata.md`. Now referenced by `contract.md` Step 5 (the GitHub `BEHIND`-masking rationale stays inline since it's load-bearing there).

Both are indexed in the `github.md` / `gitlab.md` TOCs, and the dispatch pattern is documented in `_shared/vcs/CLAUDE.md`. No pr-followup file references a GitHub and GitLab recipe side-by-side anymore. `verify:skill-deps` (no reverse deps) and `verify:plugin` pass; full vitest suite green (833 passed).

PR #362, opened 2026-08-04
fix(pr-followup): ride through GitHub API outages instead of dying after a few failed fetches

stan4git:
The detached watch loop re-reads gh's OS keyring on every poll. A background Git Bash process can lose that access mid-life, surfacing as silent empty fetches that march the loop to its failure cap and stop the watcher — hit three times in one session. With the out-of-session watchdog gone (removed 2026-07-29), recovery then waits for reconcile at the next session start.

- **pr-watch-guards.sh**: `watcher_pin_token` exports `GH_TOKEN` from the CLI at arm time (while the session still holds keyring access), so every poll authenticates from the environment. Honors an existing token; no-op when none is obtainable. New `MUGGLE_PR_WATCH_MAX_FETCH_FAILURES` (default 8, was a hardcoded 5).
- **arm-watcher.md**: the loop pins the token at startup, retries a failed fetch once before counting it toward the cap, and logs the error to `watch-fetch.log` so a persistent failure is diagnosable.
- **+4 guard tests** (pin exports / no-op / honors existing, constant default).

Verified: `bash -n`, vitest 14/14, eslint clean, check-skill-deps OK (342 links). E2E = SKIPPED (skill + bash tooling, no UI surface).


plugin/skills/muggle-pr-followup/arm-watcher.md
```diff
@@ -13,18 +13,21 @@ How an orchestrating session starts the watch on one PR. Every arming point runs
    ```sh
    GUARDS="<abs>/scripts/pr-watch-guards.sh"
    [ -f "$GUARDS" ] && . "$GUARDS" || exit 0   # guards gone (plugin moved/upgraded) → a newer version's watcher owns this now
+   watcher_pin_token                           # capture GH_TOKEN now, while the session still holds keyring access
    echo "$$" > "<slot>/watch.pid"
    started=$(date +%s)
    ```
 
+   `watcher_pin_token` exports `GH_TOKEN` so every poll authenticates from the environment: a detached loop that re-reads gh's OS keyring on each call can lose that access mid-life (background processes do), which surfaces as silent empty fetches that march the loop to its failure cap. Each poll's fetch is retried once before it counts against the cap, and its error is appended to `<slot>/watch-fetch.log` so a persistent failure is diagnosable rather than a bare exit.
```
stan4git: isnt this a security leak?

stan4git: Fair flag, and you were right on two counts. It's not written to disk or logged, but exporting a token into a long-lived detached process is a real smell (readable via `/proc`, inherited by children, prone to leaking into crash dumps or a future `set -x`) — and more to the point, it was a **guess**. The loop swallowed the fetch error, so I never actually knew why watchers were dying.

The error-logging (the one part worth keeping) then captured it live: `dial tcp …:443: connection attempt failed` / `error connecting to api.github.com` — a ~8-minute **network outage**, not keyring loss. The token pin was irrelevant; the watcher died with it.

So I've dropped the pin entirely and refocused this PR on the actual cause: the loop now **rides through outages** — `MUGGLE_PR_WATCH_MAX_FETCH_FAILURES` raised to 60 with a capped linear back-off, a successful fetch resets the counter, terminal exit stays reserved for MERGED/CLOSED. `78bc2c57`.

PR #365, opened 2026-08-05
[UNIT-ONLY] Capture E2E run instructions for reuse

stan4git:
## Goal

Add a Decide-phase stage to `muggle-test-prepare` that asks how this stack is run locally — startup order, manual steps, gotchas — and persists it to `~/.muggle-ai/e2e-instructions/<key>.md` so later runs and other Muggle skills reuse it instead of re-deriving.

## Why

`prepare-plan.json` already records *what to start* (name, dir, command, port). It has no field for the tribal knowledge that makes a healthy stack look broken: that the UI 500s until the API is listening, that the first build takes four minutes and is not hung, that the port is not the framework default. Every run re-derived that, or missed it.

## Changes

**New stage** — `plugin/skills/muggle-test-prepare/steps/e2e-instructions.md`. Runs after `identify-services`, since startup order is unanswerable before the service set is known. Multi-select categorisation, then one free-text turn. "Nothing special" writes a sentinel so the question is not re-asked every run.

**Machine-local storage.** The instructions live under `~/.muggle-ai/e2e-instructions/<key>.md` and never inside the user's project — a project directory is shared, versioned, and cloned onto machines set up differently, none of which is true of a local run recipe. `<key>` is the stack identity the global prepare plan already keys on.

**VCS- and OS-agnostic.** The stage invokes no version-control tool: its location is keyed on a working directory, not a checkout, so it works outside a repo and under any VCS. The key derives from the resolved absolute path with no assumption about which character separates segments.

**Scope boundary.** The plan stays the single source of truth for start commands; the markdown never restates one and points at the plan instead. It carries only what JSON has no field for — ordering, multi-step instructions, gotchas.

**Reuse rides `reusePreparePlan`** — no new preference key. This content goes stale for the same reason the service plan does, so one answer governs both. `reuse-plan.md` loads the companion and carries it forward; the short-circuit path captures it once if missing.

**Secrets.** Pointers only — an env-var or secret name, never a value; anything credential-shaped is dropped from free text. Plaintext notes are not a secret store, and this is the file someone pastes when asking why their stack will not start.

**Consumers.** `_shared/resolve-e2e-validation-context.md`, `muggle-test/execute-local.md`, and `muggle-test-feature-local/SKILL.md` read the file. The shared procedure references it by path, not by markdown link — `muggle-test-prepare` already links into `_shared/`, so a link back would close a cycle `check-skill-deps.mjs` rejects.

## Lints

`src/test/skills/e2e-instructions-contract.test.ts` pins the four properties that fail silently if the prose drifts: no document may route the file into the user's repository; the stage may not invoke a version-control tool; it may not assume a platform path separator; and the shared procedure may not markdown-link into `muggle-test-prepare/`. Includes an anti-vacuous guard.

## Validation

Strategy is `unit-only` — skill contracts are markdown and the new test is a static prose lint, so there is no runtime surface to drive a browser against. The full vitest suite, `check-skill-deps.mjs`, `typecheck`, `lint:check`, and `verify:plugin` were all run locally and are green; lint reports no errors, with 17 warnings that pre-exist in `internal/*-gate-eval/src/run.ts` and are untouched here.

### DIFF
--- a/plugin/skills/_shared/resolve-e2e-validation-context.md
+++ b/plugin/skills/_shared/resolve-e2e-validation-context.md
@@ -25,6 +25,7 @@ Resolve without prompting; use as questionnaire defaults:
 3. Candidate projects — `muggle-remote-project-list`, ranked against the repo's dev URL and the PR title.
 4. Existing test-user secrets — `muggle-remote-secret-list` per candidate project (`managed_profile_email` / `managed_profile_password`).
 5. Auth0 tenant for local dev — grep the repo env file for `*AUTH0_DOMAIN*`.
+6. E2E run instructions — read `<repo>/.muggle-ai/e2e-instructions.md` (else the sanitized-parent-dir file under `~/.muggle-ai/e2e-instructions/`) when it exists. It records startup order, manual steps, and local gotchas for this stack. Treat it as read-only input: surface its gotchas rather than re-deriving them, and never re-ask what it already answers. Absent is normal — proceed without it.
 
 ## Questions
 
--- a/plugin/skills/_shared/resolve-e2e-validation-context.md
+++ b/plugin/skills/_shared/resolve-e2e-validation-context.md
@@ -25,7 +25,7 @@ Resolve without prompting; use as questionnaire defaults:
 3. Candidate projects — `muggle-remote-project-list`, ranked against the repo's dev URL and the PR title.
 4. Existing test-user secrets — `muggle-remote-secret-list` per candidate project (`managed_profile_email` / `managed_profile_password`).
 5. Auth0 tenant for local dev — grep the repo env file for `*AUTH0_DOMAIN*`.
-6. E2E run instructions — read `<repo>/.muggle-ai/e2e-instructions.md` (else the sanitized-parent-dir file under `~/.muggle-ai/e2e-instructions/`) when it exists. It records startup order, manual steps, and local gotchas for this stack. Treat it as read-only input: surface its gotchas rather than re-deriving them, and never re-ask what it already answers. Absent is normal — proceed without it.
+6. E2E run instructions — read `~/.muggle-ai/e2e-instructions/<key>.md` when it exists, keyed on this stack's identity. It records startup order, manual steps, and local gotchas. Treat it as read-only input: surface its gotchas rather than re-deriving them, and never re-ask what it already answers. Absent is normal — proceed without it.
 
 ## Questions
 
--- a/plugin/skills/muggle-test-prepare/SKILL.md
+++ b/plugin/skills/muggle-test-prepare/SKILL.md
@@ -50,7 +50,7 @@ All launched processes are tracked in `/tmp/muggle-test-prepare.json`:
 
 This file is **ephemeral runtime state**, not the saved recipe. The durable plan lives at `<repo>/.muggle-ai/prepare-plan.json` (or the parent-dir-keyed entry in `~/.muggle-ai/prepare-plans.json`) and is consulted in [reuse-plan](./steps/reuse-plan.md) before any other stage. The two files never merge. The `test-prepare-runner` agent writes this file during execution; the triage below and Cleanup read it.
 
-Alongside the plan sits `<repo>/.muggle-ai/e2e-instructions.md` — the prose companion written by [e2e-instructions](./steps/e2e-instructions.md), holding startup order, manual steps, and local gotchas. The plan stays the single source of truth for each service's start command; the markdown never restates one.
+The prose companion to the plan is `~/.muggle-ai/e2e-instructions/<key>.md`, written by [e2e-instructions](./steps/e2e-instructions.md), holding startup order, manual steps, and local gotchas. Machine-local data stays under the Muggle home directory, never inside the user's project. The plan remains the single source of truth for each service's start command; the markdown never restates one.
 
 **On every invocation**, check this file first. If it exists with live PIDs (verify with `kill -0`), `AskUserQuestion`:
 - Option 1: "Keep them running — skip to testing"
@@ -123,7 +123,7 @@ After a test run, the caller can re-invoke for cleanup or leave services running
 - **The user may prefer to start services themselves** — always offer that option.
 - **Never start a process the user didn't approve** — approvals are granted in Decide and travel in the plan; the agent starts nothing outside it.
 - **Never read file contents outside confirmed directories** — folder names are discoverable; file contents require explicit user selection.
-- **Never write a credential into `e2e-instructions.md`** — it lands in the user's repository, which nothing gitignores. Record the env-var or secret name, never its value.
+- **Never write a credential into the E2E run instructions** — plaintext notes are not a secret store, and this is the file a user pastes when asking why their stack won't start. Record the env-var or secret name, never its value.
 - **Never leave orphan processes untracked** — every background PID goes into the tracking file.
 - **Never kill a process the user started independently** — `external: true` survives cleanup.
 - **Never assume start commands** — verify via indicator file; confirm with user.
--- a/plugin/skills/muggle-test-prepare/SKILL.md
+++ b/plugin/skills/muggle-test-prepare/SKILL.md
@@ -50,6 +50,8 @@ All launched processes are tracked in `/tmp/muggle-test-prepare.json`:
 
 This file is **ephemeral runtime state**, not the saved recipe. The durable plan lives at `<repo>/.muggle-ai/prepare-plan.json` (or the parent-dir-keyed entry in `~/.muggle-ai/prepare-plans.json`) and is consulted in [reuse-plan](./steps/reuse-plan.md) before any other stage. The two files never merge. The `test-prepare-runner` agent writes this file during execution; the triage below and Cleanup read it.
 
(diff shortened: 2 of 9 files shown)


plugin/skills/_shared/resolve-e2e-validation-context.md
```diff
@@ -25,6 +25,7 @@ Resolve without prompting; use as questionnaire defaults:
 3. Candidate projects — `muggle-remote-project-list`, ranked against the repo's dev URL and the PR title.
 4. Existing test-user secrets — `muggle-remote-secret-list` per candidate project (`managed_profile_email` / `managed_profile_password`).
 5. Auth0 tenant for local dev — grep the repo env file for `*AUTH0_DOMAIN*`.
+6. E2E run instructions — read `<repo>/.muggle-ai/e2e-instructions.md` (else the sanitized-parent-dir file under `~/.muggle-ai/e2e-instructions/`) when it exists. It records startup order, manual steps, and local gotchas for this stack. Treat it as read-only input: surface its gotchas rather than re-deriving them, and never re-ask what it already answers. Absent is normal — proceed without it.
```
stan4git: for local machine/user level data should all be in ~/.muggle-ai/e2e-instructions/

never <repo>/

plugin/skills/muggle-test-prepare/steps/e2e-instructions.md
stan4git, line 1: VCS agnostic and OS agnostic this instruction

plugin/skills/_shared/resolve-e2e-validation-context.md
stan4git: Done — the repo path is gone entirely, not just deprioritised. Sole location is now `~/.muggle-ai/e2e-instructions/<key>.md`, keyed on the same stack identity the global prepare plan already uses. Updated across all seven documents that named it (`89c97cf`).

Two knock-on effects worth flagging:

- It removed the `git rev-parse --show-toplevel` dependency for free, which is half of the other comment.
- It invalidated my secrets rationale. I had argued pointers-only from "this lands in a repo nothing gitignores" — living outside the project makes that false. I kept the rule but re-justified it: plaintext notes are not a secret store, and this is the file someone pastes when asking why their stack won't start. Say the word if you'd rather drop the rule now that the commit risk is gone.

A lint now asserts no document can route the file back into the user's repo.

**One thing I deliberately did not touch:** `prepare-plan.json` still resolves `<repo>/.muggle-ai/prepare-plan.json` first, falling back to the home dir (`steps/reuse-plan.md:9`, `steps/readiness-report.md:46`) — as does `last-host.json` and `last-project.json`. If "should all be in `~/.muggle-ai/`" covers those too, that's a real inconsistency, but it's pre-existing behaviour well outside this diff. Happy to do it as a follow-up PR — didn't want to silently widen this one.

plugin/skills/muggle-test-prepare/steps/e2e-instructions.md
stan4git: Done (`89c97cf`).

**VCS-agnostic** — the stage keyed its location off `git rev-parse --show-toplevel`, which broke it for anyone on another VCS, or working outside a repo at all. That call is gone; the file is keyed on a working directory, which is what it was actually about.

**OS-agnostic** — I had `/` and `\` replaced by `-`, which bakes in an assumption about which character separates path segments. Now: derive the key from the resolved absolute path, replacing every separator and any drive-letter colon, explicitly noting separators differ per platform. No shell-specific recipe anywhere in the stage.

Both are pinned by lints so they can't drift back: one asserts the stage invokes no version-control tool, the other that it doesn't assume a platform separator.

plugin/skills/muggle-test-prepare/SKILL.md
```diff
@@ -50,6 +50,8 @@ All launched processes are tracked in `/tmp/muggle-test-prepare.json`:
 
 This file is **ephemeral runtime state**, not the saved recipe. The durable plan lives at `<repo>/.muggle-ai/prepare-plan.json` (or the parent-dir-keyed entry in `~/.muggle-ai/prepare-plans.json`) and is consulted in [reuse-plan](./steps/reuse-plan.md) before any other stage. The two files never merge. The `test-prepare-runner` agent writes this file during execution; the triage below and Cleanup read it.
```
stan4git: can you please make sure saving user data in ~./ rather than <repo>/

src/test/skills/e2e-instructions-contract.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    path.join(SKILLS_DIR, "muggle-test", "execute-local.md"),
+    path.join(SKILLS_DIR, "muggle-test-feature-local", "SKILL.md"),
+  ];
+
+  it("no document routes the file into the user's repository", () => {
+    for (const docPath of DOCS_NAMING_THE_FILE) {
+      // A repo-relative path here would put user-level machine state under version
+      // control and ship one developer's local recipe to everyone who clones.
+      expect(read(docPath), `repo-local path in ${path.basename(docPath)}`).not.toMatch(
+        /<repo>\/\.muggle-ai\/e2e-instructions|\$REPO\/\.muggle-ai\/e2e-instructions/,
```
stan4git: same

plugin/skills/muggle-test-prepare/SKILL.md
stan4git: Done (`4ec2e70`). The prepare plan now follows the same rule as the run instructions.

`<repo>/.muggle-ai/prepare-plan.json` is gone — the keyed entry in `~/.muggle-ai/prepare-plans.json` is the sole location. Removing that branch also deleted the second `git rev-parse --show-toplevel`, so neither resolution path needs a VCS now. Updated in `SKILL.md`, `steps/reuse-plan.md`, and `steps/readiness-report.md`.

I also dropped a hardcoded `<repo>/.muggle-ai/last-host.json` from the guardrail list. That path is the last-host MCP tool's private storage detail, and the skill restating it was a layering problem independently of where the file lives — it now reads the cache through `muggle-local-last-host-get` and lets the tool own the location.

**Two things I did not move, deliberately — both need your call:**

1. **`last-host.json` and `last-project.json`.** These aren't prose: the location lives in `packages/mcps/src/shared/last-host.ts` with its own tests and tool-registry descriptions. Moving them is a real code change *and* a silent data migration — every existing user has a populated `<cwd>/.muggle-ai/last-host.json` that would be orphaned, and they'd quietly lose their cached host. That deserves its own PR with a migration step, not a rider on this one. I left `steps/check-running.md` describing the current behaviour accurately rather than documenting a move that hasn't happened — docs that lie are worse than docs that lag.

2. **Project-scoped `preferences.json`.** `muggle-preferences/ops/configure.md` deliberately offers "This project only" as a scope, and the SessionStart hook overlays `<cwd>/.muggle-ai/preferences.json` over the global one. That's a feature, not an accident — per-project preference overrides. Removing it is a product decision, not a cleanup.

Say the word on either and I'll take them as follow-ups.

src/test/skills/e2e-instructions-contract.test.ts
stan4git: Generalised (`4ec2e70`). The assertion was scoped to the run-instructions filename; it now covers any saved state:

```ts
expect(read(docPath)).not.toMatch(/(<repo>|\$REPO)\/\.muggle-ai\//);
```

So it catches the prepare plan, the run instructions, and anything added later — no document may route saved state into the user's repository.

Added a companion lint too, since the VCS point has the same drift risk: none of the three resolution documents (`e2e-instructions.md`, `reuse-plan.md`, `readiness-report.md`) may invoke `git rev-parse`.

954 tests green.

plugin/skills/muggle-test-prepare/SKILL.md
stan4git: take them as follow ups. basically you just need to check existance of the file when missing, copy over if possible.

stan4git: Agreed — both stay out of this PR, and copy-on-missing is the right migration. It removes the objection I raised: nobody loses a cached host, because the first read after the change finds the new path absent, falls back to the old one, and copies it forward. No separate migration step, no flag day.

Recording the shape so the follow-up doesn't re-derive it:

**PR A — `last-host.json` / `last-project.json` → `~/.muggle-ai/`.** Read resolves new path → if missing, read old `<cwd>/.muggle-ai/` path → if found, write it to the new location and return it. Writes always go to the new path only. Touches `packages/mcps/src/shared/last-host.ts` and its last-project counterpart, the two tool-registry descriptions, their tests, plus the docs that describe the location (`steps/check-running.md`, `muggle-test/SKILL.md`, `muggle-test-feature-local/SKILL.md`, `muggle-test-import/SKILL.md`, `muggle-test-regenerate-missing/SKILL.md`, `preference-gates/autoSelectLocalHost.md`). Leaving the old file in place is fine — it just goes unread.

**PR B — project-scoped `preferences.json`.** Genuinely different in kind: it's a deliberate feature, and the SessionStart hook overlays project over global by design. If the intent is that user *preferences* are user-level too, that's a product decision to make explicitly rather than fold into a path cleanup — so I'd want a yes on the behaviour before touching it.

I'll open PR A once this one lands, so the shared skill docs don't conflict. Say the word if you'd rather have it stacked on this branch now.

PR #366, opened 2026-08-05
Store last-host and last-project caches in the home directory

stan4git:
Follow-up to #365, agreed in review there.

## Goal

Move the `last-host` and `last-project` caches out of the user's project and into `~/.muggle-ai/`, with a migration that loses nothing.

## Why

Both caches wrote to `<cwd>/.muggle-ai/`. They are machine-local, user-level data — a project directory is shared, versioned, and cloned onto machines set up differently, and a cached dev-server URL is none of those things. #365 moved the prepare plan and the E2E run instructions for the same reason; these were left behind because they are code-backed rather than prose, and needed a migration story first.

## Storage

One `{ version, entries }` map per cache in the Muggle home directory, keyed on the resolved working directory — mirroring the existing `prepare-plans.json` convention. Per-directory semantics are unchanged.

```
~/.muggle-ai/last-host.json     { "version": 1, "entries": { "<resolved-cwd>": { host, savedAt } } }
~/.muggle-ai/last-project.json  { "version": 1, "entries": { "<resolved-cwd>": { projectId, projectUrl, projectName, savedAt } } }
```

## Migration — copy-on-read

A home lookup that misses falls back to the in-project file and copies that entry forward. No flag day, no separate migration step, nobody loses a cached host or project. The copy-forward write is guarded: a home directory that cannot be written still returns a usable entry instead of making a read throw.

**Clear deletes the in-project file too.** This is the non-obvious part. Removing only the home entry would let the very next read migrate the old value straight back — silently resurrecting the entry the user just cleared. A test in each suite pins it.

## Shape of the change

The read/write/clear/migrate logic was identical between the two caches apart from a file name and a property name, so it lives once in `shared/cwd-keyed-cache.ts`, with types and constants split out per the `preferences.ts` precedent already in that folder.

**The public surface is unchanged** — same functions, interfaces, and constants, re-exported from `last-host.ts` / `last-project.ts`. No caller moved.

The SessionStart hook reads both caches directly to emit its `Muggle Test Last Host:` / `Muggle Test Last Project:` context lines. It now reads the home files, falling back to the in-project file **without** migrating — a hook should not write. Its node block was checked against fixture directories for the home-entry, legacy-only, and empty cases.

## Validation

Strategy is `unit-only`: this is storage plumbing with no runtime UI surface to drive a browser against. The full vitest suite, `typecheck`, `check-skill-deps`, and `lint:check` were all run locally in the worktree and are green; lint reports no errors, with 17 warnings that pre-exist in `internal/*-gate-eval/src/run.ts` and are untouched here.

Coverage was added for the behaviours that would otherwise fail silently: home-directory storage, the copy-forward migration, home-entry-beats-legacy precedence, clear removing the superseded in-project file, isolation between two working directories sharing one map, and a corrupt home file degrading to null rather than throwing.

## Note

`packages/mcps/src/mcp/tools/local/tool-registry.ts` has four pre-existing `const result` locals (lines 178, 331, 378, 747) that trip the `code-standards` hook. They are unrelated to this change and I left them alone to keep the diff honest — say the word if you want them renamed here.

### DIFF
--- a/packages/mcps/src/mcp/tools/local/tool-registry.ts
+++ b/packages/mcps/src/mcp/tools/local/tool-registry.ts
@@ -46,6 +46,7 @@ import {
   clearLastHost,
   LAST_HOST_FILE_NAME,
 } from "../../../shared/last-host.js";
+import { MUGGLE_HOME_DISPLAY_DIR } from "../../../shared/cwd-keyed-cache-constants.js";
 import {
   cancelExecution,
   executeReplay,
@@ -462,7 +463,7 @@ const preferencesSetTool: ILocalMcpTool = {
 const lastProjectGetTool: ILocalMcpTool = {
   name: "muggle-local-last-project-get",
   description:
-    "Get the cached last-used Muggle Test project for a repo (read from the entry keyed on <cwd> in ~/.muggle-ai/last-project.json). " +
+    `Get the cached last-used Muggle Test project for a repo (read from the entry keyed on <cwd> in ${MUGGLE_HOME_DISPLAY_DIR}/${LAST_PROJECT_FILE_NAME}). ` +
     "Returns the project ID, URL, name, and saved-at timestamp, or null if no cache exists. " +
     "Skills consult this when 'autoSelectProject = always' to silently reuse the project the user picked previously, instead of presenting the project picker every time.",
   inputSchema: LastProjectGetInputSchema,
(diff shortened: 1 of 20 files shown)


packages/mcps/src/mcp/tools/local/tool-registry.ts
```diff
@@ -477,7 +477,7 @@ const lastProjectGetTool: ILocalMcpTool = {
       const content = [
         "No cached last-used project for this repo.",
         "",
-        `Looked at: \`${input.cwd}/.muggle-ai/${LAST_PROJECT_FILE_NAME}\``,
+        `Looked at: \`~/.muggle-ai/${LAST_PROJECT_FILE_NAME}\` (entry for \`${input.cwd}\`)`,
```
stan4git: can you maybe make a share constant to avoid multiple copies of this path?

stan4git: Done (`369cd04`). `MUGGLE_HOME_DISPLAY_DIR` in `cwd-keyed-cache-constants.ts`, used at all nine sites — four tool descriptions and five result lines.

Worth noting what the actual duplication was: the *file names* were already constants (`LAST_HOST_FILE_NAME`, `LAST_PROJECT_FILE_NAME`), so the only part that could drift was the directory holding them, written out longhand each time. Naming that one thing fixes it; the four plain-string descriptions became template literals to use it.

The rendered strings are byte-identical, so the tool-surface snapshot is untouched — which is a decent signal the change is purely a de-duplication.

I left `~/.muggle-ai/telemetry/failure-events.jsonl` (line 655) alone — different subsystem, not a cache, and pulling it in would widen this diff past the review. Happy to fold it in if you'd rather the constant covered every home-dir mention.

PR #367, opened 2026-08-05
Resolve preferences from the home directory only

stan4git:
Follow-up to #365. Companion to #366 — see *Overlap* below.

## Goal

Resolve preferences from `~/.muggle-ai/preferences.json` only. The `<repo>/.muggle-ai/preferences.json` layer is removed.

## This removes a shipped capability — on purpose

`ops/configure.md` offered a scope question with "This project only", and the SessionStart hook overlaid project over global. That is gone. Per-repo gating is no longer possible for any knob. The ones most likely to be missed:

- **`defaultExecutionMode`** — a repo testable only in the cloud and one that runs locally now share a single default.
- **`autoSelectProject` / `autoSelectLocalHost`** — the *caches* stay per-repo (#366), so a reused project or host is still the right one for that repo. Only the ask-vs-silent choice became global.
- **`showElectronBrowser`, `autoE2ETest`** — same shape.

Flagging explicitly because it is a product decision, not a cleanup side effect.

## Migration — non-destructive, and deliberately not a merge

A project overlay holds settings chosen *for one project*. Merging it into global would apply them everywhere, which is a data-loss-shaped bug wearing a helpful hat. So only the safe case copies:

- **No global file** → the project file is copied forward. This runs *before* the first-run defaults seed; seeding first would create the global file and demote the user's only copy of their settings to "shadowed".
- **Global file exists** → nothing is written, nothing is deleted. The SessionStart hook emits a one-time notice naming the project file and the keys whose values no longer take effect, so the user can re-apply the ones they want everywhere. Keys that already agree with global are omitted — nothing changed for those.

The notice is suppressed by a stamp keyed on the project path and the diverging key set, so it re-fires only if either changes.

## Strict input schema

`PreferencesSetInputSchema` is now `.strict()`, deviating from its sibling contracts which strip unknown keys. A caller predating this change still sends `scope: "project"` — and stripping that key would write the preference to **every** project instead of the one the caller named. Rejecting surfaces the version skew; stripping silently widens the blast radius. The reason is recorded in a comment at the call site.

## Windows bug fix — separable, and worth landing regardless

Rewriting the hook exposed this. It interpolated the shell's `$HOME` into `node -e`. Under Git Bash that is `/c/Users/<name>`, which Windows node resolves to `C:\c\Users\...` and cannot open — so the block threw and printed nothing. Because the `[ -f ]` test had already passed *in bash*, the "not configured" fallback never fired either.

```
bash guard PASSES   →  [ -f /c/Users/stan4/.muggle-ai/preferences.json ]
Windows node FAILS  →  ENOENT C:\c\Users\stan4\.muggle-ai\preferences.json
os.homedir()        →  reads 23 keys
```

**No Windows user has ever had a preferences line injected into session context.** Node now resolves via `os.homedir()`.

This is a live behaviour change beyond the stated scope — Windows users' saved global preferences begin taking effect for the first time. It is in this PR because it is in a file this PR rewrites, but it stands alone and I am happy to split it out if you would rather it landed independently of the scope decision above.

## Overlap with #366

Both PRs branch from `f67e524` and both touch `tool-registry.ts`, `ensure-electron-app.sh`, the tool-surface snapshot, and `autoSelectLocalHost.md`. Whichever merges second needs a rebase; I will handle it.

## Deliberately not done

- **The hook's defaults map is a stale duplicate of `DEFAULT_PREFERENCES`** — it hardcodes mostly `ask` and omits four keys entirely, where `DEFAULT_PREFERENCES` says `always`. Syncing them would flip every unset key from `ask` to `always` for every user. That is its own decision. Consequence here: the notice compares project values against the hook's resolved values, which is self-consistent with what skills actually see.
- **`tool-registry.ts` churn.** The `code-standards` hook refused writes to that file until four pre-existing vague `result` locals were renamed. Mechanical and test-covered, but unrelated to preferences and it inflates that file's diff.

## Validation

Strategy is `unit-only`: configuration resolution with no runtime UI surface to drive a browser against. The full vitest suite, `typecheck`, `check-skill-deps`, and `lint:check` were all run locally in the worktree and are green; lint reports no errors, with 17 warnings pre-existing in `internal/*-gate-eval/src/run.ts`. `preference-gates-lint.test.ts`, which cross-checks the gate docs against the `PreferenceKey` enum and `DEFAULT_PREFERENCES`, passes.

Coverage added for global-only resolution, copy-forward when global is absent, the no-merge path when global exists, the rejected `scope` argument, and the hook notice across repeat runs.


packages/mcps/src/mcp/local/contracts/preferences-schemas.ts
```diff
@@ -23,12 +23,12 @@ export const PreferencesSetInputSchema = z
     value: z
       .enum(preferenceValueValues)
       .describe("The value. Most keys accept always/ask/never; defaultExecutionMode accepts local/remote/ask."),
-    scope: z
-      .enum(["global", "project"])
-      .default("global")
-      .describe("Write to global (~/.muggle-ai/) or project (.muggle-ai/) preferences."),
```
stan4git, line 29: why delete?

packages/mcps/src/shared/preferences-types.ts
```diff
@@ -79,10 +79,37 @@ export enum PreferenceValue {
...
 /**
- * Partial preferences — used for per-project overrides.
+ * Subset of preference keys — the shape actually stored on disk.
  */
 export type IPartialPreferences = Partial<IPreferences>;
 
+/**
+ * Result of reconciling a legacy per-project preferences file against the global one.
+ */
+export enum ProjectPreferencesMigrationOutcome {
```
stan4git: moveout

packages/mcps/src/shared/preferences-service.ts
stan4git, line 1: use a more accurate name

plugin/skills/muggle-preferences/SKILL.md
stan4git, line 1: Can you make sure you have evals to cover all cases/skills you changed?

packages/mcps/src/mcp/local/contracts/preferences-schemas.ts
stan4git: Because `scope` had exactly two values, `global` and `project`. With the project layer gone it has one legal value and therefore selects nothing — a parameter the model can still pass that cannot change any outcome. `cwd` existed only to locate the project directory for `scope: "project"`, so it has no remaining caller.

Your question did expose a weakness in how I handled it, though, and I've fixed that (`000dd99`). Deleting the keys plus bare `.strict()` produced *"Unrecognized key: scope"* — which tells a stale caller nothing about why or what to do. The message now names them:

> Expected only { key, value }: per-project preferences were removed, so scope and cwd are no longer accepted.

One trade-off worth surfacing: in Zod 4 an object-level message also covers a non-object input, so I worded it to read correctly in both cases rather than asserting something misleading when someone passes a string. Field-level errors are unaffected — a bad `value` still reports against `value`, and there's a test pinning that the object message doesn't swallow ordinary validation failures.

If you'd rather keep `scope` in the schema as an explicitly-rejected key so it self-documents in the tool surface, that's a reasonable alternative — say so and I'll switch.

packages/mcps/src/shared/preferences-types.ts
stan4git: Moved (`000dd99`) to `packages/mcps/src/shared/project-preferences-reconcile-types.ts`.

Agreed on the reasoning: `preferences-types.ts` holds the preference *vocabulary* — the keys, the values, the on-disk file shape. Reconciling a superseded per-project file is a separate and temporary concern that the vocabulary outlives, so bolting it in there would leave a permanent module carrying a transitional type.

The enum and report interface went with it, the barrel re-exports the new module, and the four consumers were repointed.

packages/mcps/src/shared/preferences-service.ts
stan4git: Renamed (`000dd99`): `migrateProjectPreferences` → `reconcileProjectPreferences`.

The tell was in the file already — the function's own doc comment opened with *"Reconcile a legacy `<cwd>/.muggle-ai/preferences.json` against global preferences"* while the name said *migrate*. And only one of its three branches migrates anything: the other two report that nothing needed doing, or that keys are shadowed and were deliberately left alone. "Migrate" promised a write that usually doesn't happen.

One canonical term now, so the enum, report interface, and the local in `setup.ts` follow: `ProjectPreferencesReconcileOutcome`, `IProjectPreferencesReconcileReport`, `projectPreferencesReconcile`.

The module header was inaccurate for the same reason — it advertised "read, write, migrate, validate" — and now reads "resolve, write, reset, validate", which is what the module actually exports.

If you meant the *module* name rather than the function, say so and I'll take another pass.

plugin/skills/muggle-preferences/SKILL.md
stan4git: Added both layers (`000dd99`), but I want to be straight about what actually gates in CI, because the answer surprised me.

**In-repo `evals/evals.json` is not wired to anything.** Only two skills have one (`muggle-pr-followup`, `muggle-test-regenerate-missing`), `muggle-test-prepare/evals/` is gitignored, and nothing in this repo reads those files — no runner, no workflow, no script. They're skill-creator artifacts you run on demand. The CI evals (`skill-gate-eval`, `skill-routing-eval`) are scenario-driven from the private brain repo. So adding an eval file alone would have looked like coverage without being coverage.

**So I added a static lint that does run in CI** — `src/test/skills/preferences-user-level.test.ts`, same Layer-1 pattern as the other skill lints. It walks every skill doc and asserts none offers a per-project scope, and none tells an agent to pass a `scope` argument the tool now rejects — that second one is the sharp edge, since a doc promising `scope` sends the agent into a call that always fails. Plus positive assertions that the skill, gate contract, and router each describe user-level resolution. Anti-vacuous guard included.

**And the eval file, per the existing convention** — `plugin/skills/muggle-preferences/evals/evals.json`, three cases: a per-project request (must refuse the scope and confirm before writing globally), the storage question (two-layer resolution), and a stale in-repo file (explain it isn't read, offer to re-apply, don't delete it).

Coverage gap I'm not closing here: the skills #366 touches (`muggle-test`, `muggle-test-feature-local`, `muggle-test-import`, `muggle-test-regenerate-missing`, `check-running`) only had a cache path corrected. Want eval coverage there too, or is a path correction below the bar?

packages/mcps/src/shared/preferences-service.ts
stan4git: the file name!

stan4git: Got it — renamed the file, not just the function. `preferences.ts` → **`preferences-service.ts`** (`27f1348`).

The reason it read wrong: sitting next to `preferences-types.ts` and `preferences-constants.ts`, a bare `preferences.ts` looks like the third member of a data triad — "the preferences" — when it is actually the service over them. Its own header already called it a service. Three importers repointed; the barrel re-exports the new path.

I did keep the bare-domain siblings alone (`auth.ts`, `data-dir.ts`, `last-host.ts`) — they aren't part of a types/constants triad in the same way, so the ambiguity doesn't arise. Say the word if you want the convention applied more broadly.

**Also rebased onto master now that #366 landed** — they conflicted in `tool-registry.ts`, as expected. The resolution keeps #366's shared-constant work and this PR's variable renames and preferences changes; the branch is one signed commit on current master, `behind_by=0`.

While resolving I found **my #366 fix was incomplete**, and fixed it here. Five result lines still spelled out `~/.muggle-ai/` longhand, because they use the backtick-escaped form inside a template literal and my replacement only matched the unescaped one. Worse, the grep I verified with searched for the literal *filename*, which those lines don't contain — so it reported zero and I believed it. They now use `MUGGLE_HOME_DISPLAY_DIR` like the rest.

PR #368, opened 2026-08-06
Make post-merge cleanup verify each step and report per step

stan4git:
## Goal

Make post-merge cleanup verify each step and report per step, so a cleanup cannot report success while a step silently didn't run.

## Why

The procedure listed three steps and asserted nothing about whether they happened. In practice a cleanup then reported success while two of them had not:

- the **remote branch survived** — the provider's auto-delete-on-merge was treated as being the step, and it fired for two PRs out of three
- the **session slot and prepare artifacts were never cleared** at all

Neither gap was visible in the output, so it stood as "cleanup complete" until someone audited it. That is the failure mode this change targets: not a step that fails loudly, but one that is skipped quietly.

## Changes

Every step now states what proves it succeeded, and the procedure ends by reporting per step from those checks rather than from the fact that a command was issued.

Three rules the old text did not have:

**Auto-delete is a setting, not a guarantee.** It can be off for the repo, off for a fork, or simply not fire. The ref is queried rather than assumed gone.

**`git branch -d` cannot pass after a squash merge.** Squashing mints a new commit carrying the same tree, so the branch tip is never an ancestor of the base. This is not a safety check failing — it is one that can never succeed — so the answer is a content check against the base before `-D`, not `-D` on faith.

**Scope is stated.** Session slots live in the home directory, not the project, and are cleared only when terminal — a slot for an open PR is live state a watcher reads. Prepare artifacts are limited to this run, since another session may be mid-prepare with services running; orphans get reported, not deleted.

Step 1 also now separates unlinking a dependency **link** from deleting a real **directory**. The two are indistinguishable in a listing and very much not in a `rm -r`.

## Reporting

The procedure ends by printing one row per step, each cell filled from that step's verification check. Markers distinguish verified-done, done-with-a-caveat, not-done-plus-reason, and nothing-to-do.

Every step gets a row. A step that did not run is never omitted, and a step whose verification did not actually happen is never marked done — a silent gap is worse than a visible one, because it is the version the user ends up believing. Anything deliberately left alone is stated below, so out-of-scope is declared rather than invisible.

See the `## Report` section of the diff for the exact shape.

## Lint

Extends `post-merge-cleanup-link-safe.test.ts` to pin the new rules, plus a structural check that every numbered step carries a `**Verify:**` line.

The existing lint earned its keep here: an earlier draft of this rewrite dropped the unlink instruction in favour of vaguer wording, and the test caught it. That is exactly the drift these prose lints exist for.

## Validation

Strategy is `unit-only` — this is a procedure document and a static prose lint, with no runtime surface to drive a browser against. The full vitest suite, `typecheck`, `check-skill-deps`, and `lint:check` all run green in the worktree; lint reports no errors, with 17 warnings pre-existing in `internal/*-gate-eval/src/run.ts`.

### DIFF
--- a/plugin/skills/_shared/post-merge-cleanup.md
+++ b/plugin/skills/_shared/post-merge-cleanup.md
@@ -4,6 +4,86 @@ Gated by [`autoCleanup`](../muggle-preferences/preference-gates/autoCleanup.md).
 
 On `always`, the steps below run as one pre-authorized sequence (no per-step prompts). Stop on the first failure; do not force.
 
-1. **Remove the worktree — link-safe.** `git worktree remove {worktreePath}`, only if a worktree was used. A worktree's dependency dir (e.g. `node_modules`) is often a **link** (symlink, or a Windows junction) to a shared tree, not a real copy — and a forced/recursive delete follows the link and wipes that shared target, breaking every worktree. So never `--force`: unlink the dependency link first (remove the link only, using the host OS's unlink), then a plain `git worktree remove {worktreePath}`.
-2. `git branch -d {branch}` — **skip when no worktree was used**: the branch is the user's current live checkout (a bootstrap/auto-track watcher), and the checked-out branch must never be deleted. Then `git push origin --delete {branch}`.
-3. Clear `.muggle-ai/` session folders for this branch's runs and stale `/tmp/muggle-prepare-*.log` files. Cloud results stay.
+**Verify every step, assume none.** Each step states what proves it succeeded. A step whose side effect *usually* happens is not a step that ran — that assumption is how a branch survives a cleanup that reported success. The [report](#report) states verified state, never intent.
+
+## Preconditions
+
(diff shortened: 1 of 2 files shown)


stan4git:
**`skill-gate-eval` went red on `2559fa4`, and it was infrastructure, not this change — re-run is green.**

The run died with `terminal_reason: "api_error"` / *"API Error: Connection closed mid-response"* while executing the `showElectronBrowser` gate against `muggle-test-feature-local`. That is the eval harness losing its connection mid-run, not an assertion failing.

This PR touches `plugin/skills/_shared/post-merge-cleanup.md` and `src/test/skills/post-merge-cleanup-link-safe.test.ts`. Neither is reachable from the `showElectronBrowser` gate or the `muggle-test-feature-local` skill, so the diff cannot influence that scenario. Every gate that actually completed passed — `checkForUpdates` at 3/3, 100% across all reps.

I re-ran the failed job rather than editing anything. Changing a diff to satisfy a transient API error would add unrelated churn and, worse, disguise a flaky eval as a code defect.

Now **12/12 green**.

PR #369, opened 2026-08-06
fix(guardrails): accept a seeded slot as the watcher hand-off

stan4git:
## The bug

The watcher-arm gate (#361) required a **live poller** — a signalable `watch.pid`, a fresh heartbeat, or `result.md` — to consider a PR handled. That made it fire on the **seeded-but-not-yet-armed** slot, which is the state a background job legitimately ends in, and the only way past it was the `MUGGLE_WATCH_SKIP` hatch.

A gate that manufactures the skip it exists to prevent is worse than no gate: it trains the caller to reach for the escape valve. That is exactly what happened on muggle-ai-brain#146 in the session that wrote this fix.

## Why liveness was the wrong bar

Watchers are **session-scoped** — the monitor and any recovery cron both die with the session that started them — so *"a poller is running right now"* is never the durable guarantee. The **slot on disk** is: reconcile re-arms every open slot whose poller went silent, and finalizes it once the PR goes terminal.

Observed end to end: **muggle-ai-brain#146** was seeded but never armed, merged on 2026-08-05, and was finalized by a later session's reconcile on 2026-08-06 (`result.md` written, state → `merged`) with **no watcher ever running**. The seeded slot did its job.

## The change

The gate now blocks only when a handled PR has **no session slot at all** — the genuine *"Stage 8 never ran, nothing will ever pick this up"* case — and accepts any slot that tracks the url, armed or not.

- `findUnarmedHandledPrs` → **`findUntrackedHandledPrs`**; the pid/heartbeat liveness probe is gone.
- `WatchGateDecision.owed` → **`.untracked`**.
- Block message now says seeding is what matters and that reconcile arms it from there, instead of demanding a live watcher.

## Tests

Pin the regression directly:
- a **seeded slot with no watcher** leaves the gate silent (blocks before seeding, silent after — asserted in the same test),
- a slot whose **watcher died** (stale pid, no heartbeat) is likewise accepted,
- a handled url with **no slot** still blocks,
- malformed / missing `prs.json` slots are ignored.

## Verification

- Full suite **981 passed**, 11 skipped, 72 files.
- `typecheck` clean, `lint:check` 0 errors, `check-skill-deps` OK (344 links).
- Rebuilt `guardrails.mjs` verified end-to-end: **blocks** with no slot, **silent** with a seeded slot.

Also renames a vague `result` binding in `reportGate` that the code-standards hook flagged on touch.

### DIFF
--- a/plugin/scripts/guardrails.mjs
+++ b/plugin/scripts/guardrails.mjs
@@ -1,4 +1,4 @@
-import { readFileSync, existsSync, mkdirSync, writeFileSync, readdirSync, statSync } from 'fs';
+import { readFileSync, existsSync, mkdirSync, writeFileSync, readdirSync } from 'fs';
 import { isAbsolute, resolve, join } from 'path';
 import { homedir } from 'os';
 
@@ -144,7 +144,6 @@ function e2eGateDecision(state, maxBlocks = MAX_E2E_BLOCKS) {
   if (blockCount >= maxBlocks) return { action: "release" /* Release */, blockCount };
   return { action: "block" /* Block */, blockCount: blockCount + 1 };
 }
-var HEARTBEAT_FRESH_MS = 15 * 60 * 1e3;
 var WATCH_SKIP_MARKER = /^\s*echo\s+["']?MUGGLE_WATCH_SKIP\b/;
 function isWatchSkipMarker(cmd) {
   return WATCH_SKIP_MARKER.test(cmd);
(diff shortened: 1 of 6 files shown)


stan4git:
Second, independent reason the liveness probe was the wrong bar — it is **structurally broken on Windows**.

`watch.sh` claims the slot with `echo "$$" > watch.pid`, which under Git Bash/MSYS is an **MSYS pid**. The gate's check was Node's `process.kill(pid, 0)`, which uses **Windows pids**. They are different number spaces for the same process:

```
watch.pid (MSYS): 539822
bash  kill -0 539822        → LIVE
Windows pid for that proc   → 94640
node  process.kill(539822,0) → ESRCH
```

Measured just now against a live, heartbeating monitor on PR #369. So on Windows the gate could never observe a live watcher: it would block on every turn end and push the caller to `MUGGLE_WATCH_SKIP` — which is exactly the failure loop this PR was opened to end, only worse than described, since it applied even to a correctly armed watcher.

Dropping the liveness probe in favor of slot existence sidesteps this entirely — no cross-runtime pid comparison remains in the gate. (The `watch.pid` lease itself is unaffected and still correct: `arm-watcher`'s dedup and the loop's supersede guard both read it from **bash**, where the pid space matches.)

PR #373, opened 2026-08-07
fix(skill-eval): retry the routing preflight instead of failing on one probe

stan4git:
## Problem

`skill-routing-eval` fires **one** preflight probe and fails the entire job if it doesn't route:

```
route="$(python … --probe 'test my changes before I open the PR' …)"
case "$route" in none|""|ERROR|TIMEOUT|THROTTLED) exit 1 ;;
```

Routing is probabilistic, so a single unlucky `none` kills a 180-minute run — and because this check is required, it blocks **every** PR that touches a skill description. The nightly on `master` has failed exactly this way since **2026-08-02** (`preflight route=none`), so the block is repo-wide, not PR-specific.

## What I ruled out

| Hypothesis | Evidence |
|---|---|
| Expired `CLAUDE_CODE_OAUTH_TOKEN` | `skill-gate-eval` passes on the **same** secret in the same workflow |
| Plugin fails to install | CI log: `Successfully installed plugin: muggleai@muggle-works` / `Status: ✔ enabled` |
| A skill description regressed routing | No skill frontmatter changed between the last green nightly and the first red one |
| A plugin hook/script broke | Only `plugin/.claude-plugin/plugin.json` (version bump) changed in that window; all SessionStart/UserPromptSubmit hooks exit 0 |

The probe query itself is fine — it routes to `Skill{muggle-test}` when run locally.

That leaves the probe's own nondeterminism (and the unpinned `npm install -g @anthropic-ai/claude-code`, which upgrades daily) as the moving part. Either way, one sample is not evidence the harness is down.

## Change

Probe up to `PREFLIGHT_ATTEMPTS` (default 3) times, pass on the first concrete route, fail only when every attempt fails. Each attempt is logged, so a genuine outage still looks different from a flake:

```
preflight attempt=1 route=none
preflight attempt=2 route=muggle-test
plugin loaded; routing works (route=muggle-test)
```

Behaviour is unchanged when the harness is truly broken — three failures still exit 1, now naming the attempt count.

## Verification

- YAML indentation preserved inside the `run: |` block (all lines ≥10 spaces)
- Extracted shell block passes `bash -n`
- Probe reproduced locally → `Skill{skill: muggle-test}`, confirming the query and parser work

Scoped deliberately to the preflight. If the underlying probe turns out to fail consistently rather than intermittently, the retry logs will show `route=none` on all 3 attempts and we chase the CLI pin separately.

https://claude.ai/code/session_015EzkKPGtBY821LqJ62BtVz

### DIFF
--- a/.github/workflows/skill-eval.yml
+++ b/.github/workflows/skill-eval.yml
@@ -518,6 +518,7 @@ jobs:
             || { echo "muggle plugin failed to install/enable" >&2; exit 1; }
 
       - name: Preflight — confirm the plugin actually routes
+        id: preflight
         if: steps.scope.outputs.should_run == 'true'
         run: |
           set -euo pipefail
@@ -525,20 +526,37 @@ jobs:
           # Routing is probabilistic, so one probe is not evidence the harness is
           # broken — a single unlucky `none` used to fail a 180-minute job and
           # block every PR that touches a skill description. Probe up to
-          # PREFLIGHT_ATTEMPTS times, passing on the first concrete route; only a
-          # clean sweep of failures means the plugin, auth, or session is down.
-          # A concrete route includes short aliases (mtest, mdo, …), which the
-          # probe lands on nondeterministically.
+          # PREFLIGHT_ATTEMPTS times and pass on the first concrete route.
+          #
+          # A pass that needed more than one attempt is not reported as a clean
+          # pass: the probe is unstable, and hiding that behind a green check is
+          # what lets flakiness rot silently. It does not block — it surfaces, as
+          # a workflow warning and a banner on the PR's routing comment.
           route=""
+          used=0
           for attempt in $(seq 1 "${PREFLIGHT_ATTEMPTS:-3}"); do
+            used=$attempt
             route="$(python internal/skill-routing-eval/router_eval.py \
               --probe 'run a muggle E2E acceptance test on my web app' --repo-root "$CLEAN_CWD")"
             echo "preflight attempt=$attempt route=$route"
             case "$route" in
               none|""|ERROR|TIMEOUT|THROTTLED) ;;
-              *) echo "plugin loaded; routing works (route=$route)"; exit 0 ;;
+              *)
+                {
+                  echo "route=$route"
+                  echo "attempts_used=$used"
+                  echo "unstable=$([ "$used" -gt 1 ] && echo true || echo false)"
+                } >> "$GITHUB_OUTPUT"
+                if [ "$used" -gt 1 ]; then
+                  echo "::warning title=Routing preflight inconsistently passed::Routed as '$route' only on attempt $used of ${PREFLIGHT_ATTEMPTS:-3} — the probe is unstable. Not blocking, but the routing signal on this run is weaker than a clean pass."
+                  echo "plugin loaded; routing INCONSISTENTLY passed (route=$route on attempt $used)"
+                else
+                  echo "plugin loaded; routing works (route=$route)"
+                fi
+                exit 0 ;;
             esac
           done
+          { echo "route=$route"; echo "attempts_used=$used"; echo "unstable=true"; } >> "$GITHUB_OUTPUT"
           echo "Preflight failed: '$route' on all ${PREFLIGHT_ATTEMPTS:-3} attempts. Plugin not loaded, auth missing, or probe session broken." >&2
           exit 1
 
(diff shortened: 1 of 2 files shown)


.github/workflows/skill-eval.yml
```diff
@@ -522,19 +522,25 @@ jobs:
...
-          # includes short aliases (mtest, mdo, …), which are valid muggle routes
-          # the probe lands on nondeterministically. `none`/empty means the
-          # plugin failed to load or auth was rejected; ERROR/TIMEOUT means the
-          # probe session itself is broken — every chunk probe would fail the
-          # same way, so die here instead of burning the whole eval budget.
-          case "$route" in
-            none|""|ERROR|TIMEOUT|THROTTLED) echo "Preflight failed: got '$route'. Plugin not loaded, auth missing, or probe session broken." >&2; exit 1 ;;
-            *) echo "plugin loaded; routing works (route=$route)" ;;
-          esac
+          # Routing is probabilistic, so one probe is not evidence the harness is
```
stan4git, line 526: I think the key requirement here is not trying to draw a clear line for "success/failure", but rather, presenting the nature "unstableness" to the owner, via CI result.

I suggest we make a ending status something like "inconsistently passed", that is not blocking check in, but has very clear visual/text cue in the PR that PR owner would not missed.

stan4git: Agreed — and you're right that the binary framing was the wrong goal. Done in `287887b`.

A preflight that needed more than one attempt no longer reports as a clean pass. It still doesn't block (the check stays green), but it announces itself in two places the owner can't scroll past:

**1. Workflow annotation** — surfaces on the run and in the checks UI:
```
::warning title=Routing preflight inconsistently passed::Routed as 'muggle-test' only on
attempt 2 of 3 — the probe is unstable. Not blocking, but the routing signal on this run
is weaker than a clean pass.
```

**2. Banner on the PR's sticky routing comment** — prepended to the existing `🧭 Skill routing eval` comment, so it lands where you already read results:

> ⚠️ **Routing preflight inconsistently passed** — routed only on attempt 2 of 3 (`muggle-test`). Not blocking, but the routing signal on this run is weaker than a clean pass.

The step now exports `route` / `attempts_used` / `unstable` as outputs so the report step can render that without re-deriving anything.

Three states instead of two:

| Outcome | Check | Cue |
| :-- | :-- | :-- |
| routed on attempt 1 | green | none |
| routed on attempt 2+ | green (non-blocking) | annotation + PR banner |
| all attempts failed | red | existing failure message, now naming the attempt count |

Worth flagging separately: while chasing this I found the preflight was failing for a **real** reason, not just flakiness — `021b25d` in this PR. The probe ran `'test my changes before I open the PR'` inside an empty `CLEAN_CWD`, where the model correctly answers "there's nothing here to test" and never routes; and muggle's MCP tools are deferred, so a healthy session calls `ToolSearch` first, which the parser scored as no-route. Reproduced locally with an isolated `CLAUDE_CONFIG_DIR` + credentials, where `system:init` shows the plugin loaded and the muggle commands present while the old probe still returned `none`. That's why `master`'s nightly has failed since 2026-08-02 with no repo change behind it.

stan4git: do you have followup fix item then for the problem?

stan4git: Yes — filed as #377, with the evidence rather than a hunch.

**The one that matters:** this PR fixed the *preflight* only. The 391-query sweep still uses the strict first-`tool_use` parser (`router_eval.py:157`), so it plausibly has the same blindness. Captured locally under CI-identical conditions:

```
T1 [text] "I'll test your app end to end. Let me load the muggle tools first."
T2 [TOOL] ToolSearch {"query":"muggle test e2e browser"}
```

First tool call is `ToolSearch` (muggle's MCP tools are deferred) → strict parser scores `NONE`, while `system:init` shows the plugin loaded and the `muggle:*` commands present. Every query routing through tool-loading reads as a miss.

I deliberately did **not** widen the fix into the sweep here — changing scoring for all 391 queries in a PR meant to unblock CI would mix an infrastructure repair with a change to what the eval measures, and I'd have no baseline to compare against. The clean read is: land this, let the first full sweep since 2026-08-02 complete, and see where the numbers fall. If they sit well below the 0.95 bar and cluster on muggle-tool-heavy queries, #377 is confirmed and the parser change is justified by data.

Two smaller items in #377:

- `CLEAN_CWD` is empty, so change-driven queries ("test my changes…") are unwinnable there by construction — the model correctly answers that there's nothing to test. I verified it still declines even in a seeded git repo with a real uncommitted change (*"Nothing here is E2E-testable"*).
- `npm install -g @anthropic-ai/claude-code` is unpinned, so eval numbers aren't comparable across days — which is exactly what made this failure look repo-caused when it wasn't.

To be explicit about scope: this PR is the CI repair, #377 is the question of whether the eval is measuring what we think it measures. I'd rather not answer the second one by guessing.

PR #375, opened 2026-08-11
Default the PR-watch lifetime to 7 days, configurable by preference

stan4git:
## Goal

Default the PR-watch lifetime to 7 days instead of 6 hours, and make it configurable through a `watcherLifetime` preference offering `1d`, `7d`, and `never`.

## Why

A watch loop retired itself after 6 hours, so every open PR needed re-arming four times a day. Each re-arm costs a model turn plus a drain tick — one session spent **seven** of them re-confirming byte-identical state on a PR whose only blocker was a human approval.

Polling itself is free; the detached loop costs nothing while it runs. So the fix is a longer life, not more frequent arming.

## `never` is a real trade, stated plainly

The cap is the **only time-based reaper** for a detached loop whose session has gone. `watcher_superseded` retires a loop only when a *newer* arm claims its slot, which never happens if nothing re-arms. So `never` means an orphaned loop polls until the machine restarts or someone kills it.

That is not hypothetical here — a `muggle-ai-ui-pr412` loop has been polling for ~8 days. The option is included deliberately, and the gate doc says this outright rather than implying a cap always protects.

## Shape

A duration doesn't fit `PreferenceValue`'s `always|ask|never` vocabulary, so the key carries its own values scoped through `PREFERENCE_ALLOWED_VALUES` — the pattern `defaultExecutionMode` already uses for `local`/`remote`. `never` reuses the existing member; only `1d` and `7d` are new.

The gate contract is marked **Not gated**, matching `verboseOutput`: it configures behaviour and never prompts. Its values could not be picker values in any case — the gate lint's canonical set is alphabetic, so `1d`/`7d` are unparseable as picker values by construction.

## Plumbing

The loop is plain `sh` and cannot read `preferences.json`. The value is resolved at arm time, converted to seconds, and exported as `MUGGLE_PR_WATCH_MAX_LIFETIME`. An already-set env var still wins, so an operator can pin any value without touching the preference.

`watcher_lifetime_exceeded` now reads `0` as unbounded. Left as a plain cap, `now - started >= 0` is true immediately, so every loop would exit on its first iteration — the exact opposite of what `never` means. A test pins that.

## Note — unrelated bug this PR increases exposure to

`writePreferencesFile` rebuilds `{version, preferences}` from scratch, so **every** preference write drops unknown top-level keys — `localLlm`, `disclosureShownAt`. This PR doesn't worsen the defect (the new key lives inside `preferences`), but it does give users a fresh reason to run `preferences-set`, which is what triggers the loss. Worth fixing separately; I didn't widen this PR to cover it.

## Validation

Strategy is `unit-only`: shell guard functions, preference constants, and docs, with no runtime UI surface to drive a browser against. The full vitest suite, `typecheck`, `check-skill-deps`, and `lint:check` all run green in the worktree; lint reports no errors, with 17 warnings pre-existing in `internal/*-gate-eval/src/run.ts`.

Coverage added for the new default, the env-var override beating it, unbounded handling, the seconds mapping for every allowed value, and rejection of `always`/`ask` which mean nothing for a duration.

Line endings were verified per file against master before committing — `pr-watch-guards.sh` is LF, and a CRLF shebang would fail on Linux with exit 127.


stan4git:
**`platform-compat (windows-latest)` went red on `96592cb` — pre-existing Windows flake, not this change. Re-run triggered.**

The failure is a 5-second timeout in `src/test/guardrails/hook-execution.test.ts > pr-opened: dedupes the same PR within a session`, on `windows-latest` only. macOS and Ubuntu both passed.

It isn't the diff. Master commit `c3d2840` has produced **both outcomes** on this job across separate runs — four successes and three failures — and `a26c16c` did the same. One commit yielding both results rules out any diff as the cause.

The mechanism is in the test itself:

```ts
const r = spawnSync(process.execPath, ["--import", "tsx", CLI, sub], { … });
```

Spawning Node with a `tsx` transform, on Windows, against a 5s vitest timeout — while the suite as a whole takes ~41s on that runner. It's timing-fragile rather than broken.

Nothing in this PR reaches that path: the diff is `pr-watch-guards.sh`, preference constants/types, a new seconds-mapping module, and two docs.

Worth someone raising the timeout on that test or marking it `retry` — it costs a re-run on roughly a third of Windows CI runs today. I haven't done it here because it's unrelated to this change.

Issue #377, opened 2026-08-11
skill-routing-eval: the 391-query sweep likely shares the preflight's deferred-tool blindness


stan4git:
Follow-up to #373, which fixed the **preflight** only. The same two defects plausibly affect the main sweep, and if they do the eval has been under-scoring routing rather than measuring it.

## 1. The sweep still scores only the first `tool_use` — highest priority

`detect_route` uses `parse_route_from_stream` for the 391-query sweep (`router_eval.py:157`); #373 introduced the tolerant `parse_route_probe` for the preflight alone.

Muggle ships its MCP tools **deferred**, so a healthy session commonly loads them before invoking anything. Captured locally under CI-identical conditions (isolated `CLAUDE_CONFIG_DIR` + credentials, empty cwd), for the query *"use muggle to test my app end to end"*:

```
T1 [text] "I'll test your app end to end. Let me load the muggle tools first."
T2 [TOOL] ToolSearch {"query":"muggle test e2e browser"}
```

First `tool_use` is `ToolSearch` → strict parser returns `NONE`, despite `system:init` showing the plugin loaded and the `muggle:*` commands present. Any query whose route goes through tool-loading scores as a miss.

**Check before changing anything:** the first full sweep after #373 lands will run to completion for the first time since 2026-08-02. If pass rates sit well below the 0.95 nightly bar — especially concentrated in muggle-tool-heavy queries — this is the cause, not the skill descriptions.

## 2. `CLEAN_CWD` is empty, so change-driven queries have no referent

The sweep runs in an empty dir. Queries of the form *"test my changes before I open the PR"* are legitimately unanswerable there — the model inspects, finds no repo, and says so. Verified: it declines in an empty dir, and still declines in a seeded git repo with a real uncommitted change (*"Nothing here is E2E-testable"*).

If such queries are scored as expecting a muggle route, they are unwinnable regardless of description quality. Either seed a representative web-app fixture as the eval cwd, or drop change-driven phrasings from the positive set.

## 3. `npm install -g @anthropic-ai/claude-code` is unpinned

The workflow installs the latest CLI on every run, so eval results aren't comparable across days and a model/CLI behaviour change silently moves the bar — which is what made this failure look repo-caused. Consider pinning with a deliberate bump, so a scoring shift is attributable.

## Why this matters

[...]

PR #380, opened 2026-08-11
fix(skill-eval): give the routing eval a cwd where change-driven queries have a referent

stan4git:
Follow-up to #377 item 2.

## Problem

The routing eval runs in an empty `CLEAN_CWD`. Roughly half the `muggle-test` positives are change-driven — *"validate my changes before I open the PR"*, *"test the changes I made on that preview URL"*, *"validate everything I've been working on this sprint"*. In an empty directory those are trick questions: the model inspects, finds no repo and no diff, and correctly answers that there is nothing to test.

Traced under CI-identical conditions (isolated `CLAUDE_CONFIG_DIR` + credentials, empty cwd):

```
T1 [text] "I'll start by looking at what's actually in the working directory and what changes exist to test."
T2 [TOOL] Bash {"command":"pwd && ls -la"}
T3 [TOOL] Bash {"command":"git status …"}
T5 [text] "The working directory isn't a git repo and has no code…"
```

First `tool_use` is `Bash`, so the query scores `none` — regardless of how the skill is described.

## Evidence this is the dominant factor

Same description, same harness, only the cwd changed:

| cwd | accuracy |
| :-- | :-- |
| empty (today) | 6/26 = 23.1% |
| seeded with a changed web app | **15/26 = 57.7%** |

For contrast, wording changes moved almost nothing on the same 6 hard queries: three variants scored at or below baseline, including one using near-verbatim copies of the failing queries.

## Change

Seed `CLEAN_CWD` with a minimal web app (checkout + login) plus two uncommitted edits, so "my changes" refers to something. It's still a throwaway dir isolated from this repo, so the plugin remains the only skill source and the original reason for `CLEAN_CWD` is preserved.

## Note on the threshold

`--fail-under 100%` has not been met since 2026-08-02 — the broken preflight masked every run. On today's harness `master`'s own description scores 3/26. Once this lands, `--fail-under` should be re-baselined against `master` so the bar reflects the harness as it now behaves; that's tracked in #377.

https://claude.ai/code/session_015EzkKPGtBY821LqJ62BtVz

### DIFF
--- a/.github/workflows/skill-eval.yml
+++ b/.github/workflows/skill-eval.yml
@@ -517,6 +517,53 @@ jobs:
           claude plugin list | grep -Fq "muggleai@muggle-works" \
             || { echo "muggle plugin failed to install/enable" >&2; exit 1; }
 
+      # The eval cwd used to be an empty dir, which made every change-driven
+      # query ("test my changes", "validate what I've been working on") a
+      # trick question: the model inspects, finds no repo and no diff, and
+      # correctly answers that there is nothing to test — scoring `none` no
+      # matter how the skill is described. Seed a small web app with real
+      # uncommitted changes so those queries have a referent. It stays isolated
+      # from this repo, so the plugin is still the only skill source.
+      - name: Seed the eval cwd with a changed web app
+        if: steps.scope.outputs.should_run == 'true'
+        run: |
+          set -euo pipefail
+          mkdir -p "$CLEAN_CWD/src"
+          cd "$CLEAN_CWD"
+          git init -q
+          git config user.email eval@example.com
+          git config user.name "routing eval"
+          cat > package.json <<'JSON'
+          { "name": "shop", "version": "1.0.0", "scripts": { "dev": "vite" } }
+          JSON
+          cat > index.html <<'HTML'
+          <!doctype html><title>Shop</title><div id="app"></div>
+          <form id="login"><input name="email"><input name="password" type="password"><button>Sign in</button></form>
+          HTML
+          cat > src/checkout.js <<'JS'
+          export function total(items) { return items.reduce((n, i) => n + i.price, 0); }
+          JS
+          cat > src/login.js <<'JS'
+          export function submit(form) { return fetch("/api/login", { method: "POST", body: new FormData(form) }); }
+          JS
+          git add -A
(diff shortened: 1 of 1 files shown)


.github/workflows/skill-eval.yml
```diff
@@ -516,6 +516,53 @@ jobs:
           claude plugin list
           claude plugin list | grep -Fq "muggleai@muggle-works" \
             || { echo "muggle plugin failed to install/enable" >&2; exit 1; }
+
+      # The eval cwd used to be an empty dir, which made every change-driven
+      # query ("test my changes", "validate what I've been working on") a
+      # trick question: the model inspects, finds no repo and no diff, and
+      # correctly answers that there is nothing to test — scoring `none` no
+      # matter how the skill is described. Seed a small web app with real
```
stan4git, line 524: This is bad pattern.

Simply exclude empty should do the work. why complicate this?

stan4git: I tested it — "exclude empty" doesn't move the number. Same description, same harness, 3 runs each:

| eval cwd | accuracy |
| :-- | :-- |
| empty (today) | 6/26 = 23% |
| **real repo checkout** (just don't use an empty dir) | **7/26 = 27%** |
| minimal 2-file fixture | 5/26 = 19% |
| full web-app fixture (this PR) | **15/26 = 58%** |

Emptiness isn't the blocker — *applicability* is. `muggle-ai-works` is a CLI/plugin repo with no web surface, so on a non-empty checkout the model still declines: it answers "nothing here is E2E-testable" instead of routing. I saw the same when I seeded a plain JS repo with a real uncommitted change.

The queries assume a **web app with changes** — "didn't break checkout or login", "on that preview URL", "the flows still work". Only a cwd matching that assumption moves the score, and the minimal fixture shows it isn't fixture-size that matters, it's whether the fixture actually looks like the app the queries describe.

Happy to go the other way if you'd prefer: drop the ~12 change-driven queries from the `muggle-test` positive set instead. That's a smaller diff and no workflow shell to maintain — but it stops the eval measuring the skill's primary trigger ("test my changes before I push"), which is the phrasing real users lead with. Between a fixture nobody reads and an eval that no longer covers the main entrance, I'd rather keep the coverage — but it's your call and I'll switch if you say so.

PR #383, opened 2026-08-12
fix(pr-followup): scope watcher recovery to its owning session, and ship the watch loop

stan4git:
Two defects in `muggle-pr-followup`, found together. Independent, but one cause: the watcher's durable state recorded **what** was being watched, never **who was watching it or how**.

Root-cause write-up: `muggle-ai-brain/architecture/2026-08-12-pr-watcher-session-ownership-and-loop-drift-design.md`.

## Defect 1 — recovery had no owner

`reconcile.md` enumerated every slot under `~/.muggle-ai/muggle-do/sessions/` and re-armed each open one whose beacon was stale. No slot recorded the Claude Code session that armed it, so "is this mine?" was unanswerable. Two entry points reached that sweep with no PR named by the user — `auto-track.md` Step 0 on a bare no-arg invocation, and the `SessionStart` nudge — so following the documented procedure correctly made a session adopt every watcher on the machine, then poll and push to PRs it had no context for.

The hook compounded it by counting `*.stopped` slots — the owner's kill switch. On the reporting machine 11 of the 21 it counted were slots that had been explicitly stopped.

**Fix.** Slots carry `owner.json` (`session_id` + `claimed_at`), written by `arm-watcher.md` at arm time. Reconcile partitions owned vs foreign and re-arms only owned; finalizing a terminal PR stays ownership-free, since it resumes no watch and reads no review. New `adopt <slug>` is the only way a slot changes hands — one slug, explicit, refuses `*.stopped`. Absent `owner.json` and unset `$CLAUDE_CODE_SESSION_ID` both read as *not mine*, so nothing is adopted and no migration runs.

## Defect 2 — the loop was re-derived on every arm

`arm-watcher.md` told each arming session to *write* `watch.sh` from prose; only the guards shipped. Derivations drifted, and a dropped wake has no failure mode anyone notices — the watcher stays alive, heartbeats, logs, and never fires.

| Slot | behind-base wake | conflict wake |
|------|------------------|---------------|
| `works-pr379` | present | present |
| `works-pr381` | present | present |
| `works-pr382` | **absent** | present |
| `works-pr383` | **absent** | present |

Two of four implement only the conflict half; the `pr382` and `pr379` loops differ by 209 lines. A behind-but-clean branch reports `MERGEABLE`, so a conflict-only wake never fires for it — and `mergeStateStatus` cannot substitute, because `BLOCKED` (review required) masks `BEHIND` on any PR awaiting approval. This PR itself sat `behind_by=2` on a base requiring up-to-date heads, unmergeable, under a watcher reporting itself healthy.

**Fix.** The loop ships: `pr-watch-events.sh` holds the wake conditions as pure functions (arguments in, one event line out, no I/O — which is what makes them testable), and `pr-watch-loop.sh` is the I/O around them. `arm-watcher.md` runs it and states that authoring a per-slot `watch.sh` is prohibited. `watch_wake_rebase` carries conflicting **and** behind in one function; behind is measured with a `compare` call made only when the head/base pair moved.

## Verification

- Suite green: 1116 tests over 78 files, 21 skipped.
- `pr-watch-events.test.ts` — 20 behavioral tests, one per wake condition and its floor semantics.
- `reconcile-stale-watchers.test.ts` — the ownership partition against fixture state dirs: owned vs foreign, absent-owner-is-foreign, no-session-id fails closed, `*.stopped` ignored, finalized ignored, silence when clean.
- The behind-base test was checked against the regressed lineage to confirm it discriminates: shipped condition emits `branch behind base by 2`, the conflict-only derivation stays silent.
- Eval cases 20–24 cover owner-gated re-arm, fail-closed sweep, explicit single-slot adopt, arming via the shipped loop, and the behind-base wake.
- `check-skill-deps` clean (347 links, no cycles); typecheck and eslint clean; `bash -n` clean; `.sh` files LF.


plugin/hooks/README.md
```diff
@@ -47,7 +47,11 @@ Each wrapper short-circuits in shell first, so the common case never pays Node c
...
+`muggle-pr-followup` watchers are session-only (a monitor or `/loop` cron); they die with their session, leaving open PRs with no live poller. This is by design — a review is addressed only inside a session that carries the context to address it. The skill's [`reconcile`](../skills/muggle-pr-followup/reconcile.md) procedure recovers them at the next session start — finalizes slots whose PR went terminal, sweeps orphan crons, re-arms silently-stopped open watchers — but re-arming needs Claude tools a shell hook can't call. So this hook nudges rather than acts.
```
stan4git, line 50: why all these nitty gritty here?? move them out

stan4git: Fair — cut it back to what the hook does.

The ownership mechanics, the fail-closed rule, and the `*.stopped` / `result.md` exclusions are reconcile's contract rather than the hook's, so stating them here duplicated the skill doc and put the second copy where it would drift first. The entry now says the nudge counts only slots this session owns and reports the rest as orphans it will not act on, and links to `reconcile.md` for the rules behind it — three paragraphs down to one.

eb91e79

PR #384, opened 2026-08-12
Prepare learns the recipe once, then replays it

stan4git:
## Goal

`muggle-test-prepare` learns how to prepare this environment on its first run, then replays that recipe on every later run without re-asking.

## The problem

Preparation asked the same questions every time. It captured what the user volunteered **up front** — the moment they're least able to articulate it — and threw away everything the run itself proved:

- `smoke-test` diagnosed a failure with a concrete signal, the user picked a fix, it worked. Recorded nowhere; the next run rediscovers it and asks again.
- `start-services` waited out a slow boot. "First build takes ~4 min, don't kill it" is literally the example in the written format, sitting in the log tail, discarded.
- Services skipped with a reason, stacks needing a fresh install — both real properties, neither remembered.

## The lifecycle

**Learning run** — no recipe exists:

1. **Ask where it may read.** Scan a folder the user names, to a depth they set (2 levels by default), *or* skip the scan and take their list of paths and services.
2. **Scan** workspace manifests — `pnpm-workspace.yaml`, `workspaces`, `turbo.json`, `docker-compose.yml`, `Procfile` — to derive services, directories, start commands, ports. Never a framework default: an undeclared port is unknown, not `:3000`.
3. **Interview** only what a scan *cannot* know — which services this user needs, what can't run locally, ordering no `depends_on` declares, steps done by hand.
4. **Prepare for real**, resolving problems autonomously where safe, escalating where not, recording each signal with the fix that cleared it.
5. **One gate at the end** showing what the run actually did, asking whether to remember it.

**Replay run** — a recipe exists: execute it. No scan, no interview, no prompts. A failed step consults recorded resolutions *first*, so a problem solved once is solved the same way again in silence.

## What keeps this from decaying back into an interrogation

**Hard block is defined narrowly**: a failure the recorded resolutions don't cover, which autonomous attempts can't clear. Anything the recipe already handles isn't one, however loudly it fails; neither is a transient that clears on retry. The bar is deliberately high — every re-prompt undoes the learning.

**Nothing persists before the gate.** Declining leaves no recipe and the next run learns again. A recipe the user didn't agree to is one they have to undo; a recipe for a preparation that failed is worse than none.

**A transient is never recorded.** Nothing was learned; it was a blip. Neither is a failure that was never cleared — that becomes an exclusion with its reason, so the next run doesn't retry something known hopeless.

## Reading is granted, never assumed

The previous boundary forbade reading file contents anywhere the user hadn't hand-picked, which a scan can't work under. Rather than quietly widening it, the scan now **asks**: name a folder and a depth, or hand over the list and nothing else gets read. The granted scope is stored with the recipe so the permission isn't re-requested every run; widening it needs a fresh ask.

## Attribution

Machine-observed resolutions live in their own dated section. The sections above it are the user's own words and are never rewritten by a run — so a write-back can't clobber what they typed, and they can tell at a glance which is which.

## Validation

Strategy is `unit-only`: this is skill procedure and static prose lints, with no runtime surface to drive a browser against. The full vitest suite, `typecheck`, `check-skill-deps`, and `lint:check` all run green in the worktree; lint reports no errors, with 17 warnings pre-existing in `internal/*-gate-eval/src/run.ts`.

Twelve new lints extend `e2e-instructions-contract.test.ts`, each pinning a point where the lifecycle would collapse back into re-interrogating the user: permission asked before reading, scope confined to the grant, the scan's findings not re-asked, ports never guessed, exactly one gate, declining writing nothing, recorded resolutions consulted before prompting, transients and unresolved failures refused, observed entries kept separate from the user's words, and credentials still barred from the write-back path.

Line endings were verified per file against master before committing — this repo is mixed, and a text-mode write silently rewrites a whole file.


plugin/skills/muggle-test-prepare/SKILL.md
```diff
@@ -73,22 +73,31 @@ Gates run per [`preference-gates/README.md`](../muggle-preferences/preference-ga
...
 | # | Stage | Summary |
 |:--|:------|:--------|
-| 0 | [reuse-plan](./steps/reuse-plan.md) | Reuse saved prepare plan (gated); on reuse, skip straight to dispatch |
-| 1 | [rebase-check](./steps/rebase-check.md) | Rebase onto default branch (gated) |
-| 2 | [scope](./steps/scope.md) | Frontend / backend / full stack |
-| 3 | [viability-check](./steps/viability-check.md) | Exclude services that can't run locally |
-| 4 | [identify-services](./steps/identify-services.md) | Pick required services + startup mode |
-| 5 | [e2e-instructions](./steps/e2e-instructions.md) | Capture startup order, manual steps, local gotchas (gated); persisted for reuse |
+| 0 | [replay-or-learn](./steps/replay-or-learn.md) | Recipe exists → replay; none → learn (gated) |
+| 1 | [scan-structure](./steps/scan-structure.md) | Derive the service graph from workspace manifests |
```
stan4git: why optional step details here ???
This should be referenced under replay or learn

plugin/skills/muggle-test-prepare/SKILL.md
```diff
@@ -73,22 +73,31 @@ Gates run per [`preference-gates/README.md`](../muggle-preferences/preference-ga
...
+| 2 | [rebase-check](./steps/rebase-check.md) | Rebase onto default branch (gated) |
+| 3 | [scope](./steps/scope.md) | Frontend / backend / full stack |
```
stan4git: let's use a more accurate name here for the stage and file

plugin/skills/muggle-test-prepare/steps/confirm-recipe.md
```diff
@@ -0,0 +1,56 @@
+# Final stage — Confirm the recipe
```
stan4git: no point of adding "order" dependency here saying "final stage". Please scan all docs and make sure they don't take such dep at all.

plugin/skills/muggle-test-prepare/SKILL.md
stan4git, line 1: move optional details out

plugin/skills/muggle-test-prepare/SKILL.md
stan4git: Moved (`0d61cfd`). The table rows are one-line summaries now; everything about *what a replay may do* lives in `replay-or-learn.md`, which owns that behaviour.

Also dropped the `(gated)` markers from the rows. Which gate applies, and what each value does, is already in the step file and in the Preferences table above — a third copy in the index is the one that goes stale.

plugin/skills/muggle-test-prepare/SKILL.md
stan4git: Renamed to **`derive-service-graph`** (`0d61cfd`), file and stage.

`scan-structure` said what it looked at without saying what it produced — and "structure" is exactly the vague noun that doesn't survive contact with a second reader. The stage exists to yield a **service graph**: names, directories, start commands, ports. `identify-services` then confirms that graph with you, so the pair now reads as produce → confirm rather than two similarly-vague scan-ish names.

plugin/skills/muggle-test-prepare/steps/confirm-recipe.md
stan4git: Done, and I scanned all of them as asked (`0d61cfd`).

Four titles named their slot — and your point had already come true before you made it: adding the scan pushed `e2e-instructions` from stage 5 to 6, which broke a lint that had pinned the number. The position was load-bearing in a test, in a file that has nothing to do with ordering.

```
# Final stage — Confirm the recipe   →  # Confirm the recipe
# Stage 5 — E2E run instructions     →  # E2E run instructions
# Stage 0 — Replay or learn          →  # Replay or learn
# Stage 1 — Scan the structure       →  # Derive the service graph
```

The body prose had the same coupling, so that changed too — a prerequisite is now stated as **data**, not as order:

> ~~Runs after [identify-services]~~ → Needs the resolved service list from [identify-services]: startup order and per-service gotchas are unanswerable until the service set is known.

That stays true however the stages end up ordered, and it names the actual dependency rather than a neighbour. Same treatment for `confirm-recipe` ("needs a verified-ready environment") and the Decide table, which no longer has an ordinal column at all.

Two lints stop it creeping back: no step title may match `Stage N` / `Final stage` / `Step N`, and no step may open a line with `Runs after`.

**One I did not touch, deliberately.** `reuse-plan.md` and `check-running.md` describe the *short-circuit path* — which stages get skipped when a plan is reused. That is genuine control flow rather than a positional label, and flattening it would lose real information. If you want that restructured too, say so and I'll take it as its own change rather than smuggling it into this one.

plugin/skills/muggle-test-prepare/SKILL.md
stan4git: Thinned (`0d61cfd`). The Workflow section went from three paragraphs of per-stage behaviour plus an eight-row annotated table to: one line saying the skill learns then replays, one line contrasting the two modes, and a bare table of stage → summary.

Everything cut already existed in the file that owns it — what a replay may do (`replay-or-learn`), which gates fire (the step files plus the Preferences table), what the gate persists (`confirm-recipe`). The index kept a second copy, and a second copy is the one that drifts.

Also caught while doing this: the rename left `scan-structure.md` behind on the branch, so both it and `derive-service-graph.md` shipped describing the same stage. The commit builder reads `git diff --name-status`, where a rename is a single `R100 old new` row, and it only emitted the new path as an addition — the old path was never deleted. Fixed in `e4f732b`, and the builder now emits a deletion for the source side of a rename.

PR #388, opened 2026-08-12
ci(evals): fall back to a spare subscription token when the primary fails

stan4git:
## Problem

All three `skill-eval` jobs authenticated from a single `CLAUDE_CODE_OAUTH_TOKEN`. When that token expired or got throttled, every job's auth preflight failed and the whole suite went red until someone regenerated the secret by hand — with `skill-gate-eval` and `skill-routing-eval` each burning up to their 180-minute budget's worth of setup first.

## Change

Resolve the token instead of asserting it. A shared composite action, `.github/actions/resolve-claude-token`, probes the primary with `claude -p`; on any non-zero exit it retries with the optional `CLAUDE_CODE_OAUTH_TOKEN_2` secret and exports whichever one authenticated as `CLAUDE_CODE_OAUTH_TOKEN` for the rest of the job.

Defined once and called from all three jobs, replacing the two duplicated preflight blocks.

### Design notes

- **Any probe failure triggers the handoff, not just rejection.** A throttled primary and an expired one are indistinguishable from the caller's side, and both are reasons to use the spare.
- **Falling back warns, it does not pass quietly.** A `::warning::` annotation names the dead primary — the run survives, but the token still needs regenerating before the spare expires too.
- **Resolution is per-run and stateless.** Nothing is written back to the repo secrets, so a repaired primary is picked up again on the next run with no manual flip. Persistently repointing "token 2 is now active" would need a PAT with `Variables: write` (the existing `BRAIN_REPO_TOKEN` is scoped to the brain repo) and would strand the suite on the spare after the primary was fixed.
- **The routing job resolves before the plugin install**, so its `claude plugin marketplace add` / `install` calls run on the same credential as the eval itself.

## Setup required

Add a `CLAUDE_CODE_OAUTH_TOKEN_2` repo secret holding a second token from `claude setup-token` (no trailing newline). Until it is set, the action behaves exactly like the primary-only preflight it replaces, so this is safe to merge before the secret exists.

A spare from a *different* Claude account also raises the ceiling on concurrent eval capacity; a spare from the same account shares one quota and only buys resilience against expiry and revocation.

## Verification

The action's shipped `run:` body was extracted from the YAML and exercised against a stubbed `claude` across six cases — healthy primary, dead primary with a working spare, dead primary with no spare configured, both dead, unset primary, and single-token operation. All six behave as intended: the right slot wins, the token is exported, the warning fires only on handoff, and both no-usable-token cases exit non-zero.

That testing caught one real defect, now fixed: the unset-primary path exited `127` through a `${VAR:?}` guard and emitted a bare shell message with no GitHub annotation, unlike every other failure path in the action.

Both YAML files parse. Line endings are preserved — `core.autocrlf` is `false` and `skill-eval.yml` is CRLF in the repo, so the commit was written from raw bytes rather than normalized text.

https://claude.ai/code/session_017kGwCqxeyTggaUNbLJeDXn

### DIFF
--- a/internal/skill-gate-eval/README.md
+++ b/internal/skill-gate-eval/README.md
@@ -80,8 +80,9 @@ It checks out `muggle-ai-brain` for the scenarios, so
 CI needs two repo secrets: `CLAUDE_CODE_OAUTH_TOKEN` (subscription auth, from
 `claude setup-token`) and `BRAIN_REPO_TOKEN` (read access to the scenario repo).
 
-`CLAUDE_CODE_OAUTH_TOKEN_2` is an optional third secret holding a spare
-subscription token. Every eval job probes the primary before it starts work and
+`CLAUDE_CODE_OAUTH_TOKEN_2` is an optional secret holding a **second
+subscription token**, spare to the one above — `BRAIN_REPO_TOKEN` is repo-read
+auth and unrelated. Every eval job probes the primary before it starts work and
 switches to the spare for that run when the primary is rejected or throttled,
 so one expired token no longer red-lines the suite. The fallback raises a
 workflow warning rather than passing silently — the primary still needs
--- a/internal/skill-gate-eval/README.md
+++ b/internal/skill-gate-eval/README.md
@@ -79,3 +79,11 @@ for de-risking a refactor that changes how skills run, not their definitions.
 It checks out `muggle-ai-brain` for the scenarios, so
 CI needs two repo secrets: `CLAUDE_CODE_OAUTH_TOKEN` (subscription auth, from
 `claude setup-token`) and `BRAIN_REPO_TOKEN` (read access to the scenario repo).
+
+`CLAUDE_CODE_OAUTH_TOKEN_2` is an optional third secret holding a spare
+subscription token. Every eval job probes the primary before it starts work and
+switches to the spare for that run when the primary is rejected or throttled,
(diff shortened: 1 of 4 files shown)


internal/skill-gate-eval/README.md
```diff
@@ -79,3 +79,11 @@ for de-risking a refactor that changes how skills run, not their definitions.
 It checks out `muggle-ai-brain` for the scenarios, so
 CI needs two repo secrets: `CLAUDE_CODE_OAUTH_TOKEN` (subscription auth, from
 `claude setup-token`) and `BRAIN_REPO_TOKEN` (read access to the scenario repo).
+
+`CLAUDE_CODE_OAUTH_TOKEN_2` is an optional third secret holding a spare
```
stan4git: second, why third???? brain repo token is different???

stan4git: Fixed in ccda7ed — you're right, that was miscounted.

I was numbering *repo secrets* (`CLAUDE_CODE_OAUTH_TOKEN`, `BRAIN_REPO_TOKEN`, then the new one = third), but `BRAIN_REPO_TOKEN` is read access to the scenario repo and has nothing to do with Claude auth. Counting it in the same sequence made a secret named `_2` read as the third of its kind.

Now reads: "an optional secret holding a **second subscription token**, spare to the one above — `BRAIN_REPO_TOKEN` is repo-read auth and unrelated."

PR #389, opened 2026-08-12
fix(vcs): sign posted bodies through a shipped script, not from prose

stan4git:
## Problem

Every PR/MR body, comment and reply is supposed to end with the Muggle Works signature. Nothing appended it — the line lived in prose and was retyped at each call site, so reliability tracked how each recipe happened to state it:

| Tier | Where | Behaviour |
|------|-------|-----------|
| Shell literal | `agents/visual-walkthrough-builder.md` | Never dropped |
| Literal in a template | `do/per-comment-replies.md`, `output-templates/*` | Usually survived |
| Link to `post-signature.md` | `vcs/github/{reply-line-comment,top-level-comment,pr-edit}.md` | Model had to follow a link and reproduce an exact string — dropped here |
| Nothing at all | all four `vcs/gitlab/*` recipes, `pr-followup-helpers/reply-routing.md` | Could never be signed |

Two amplifiers: no CI or hook checked, and `muggle-pr-followup` runs on `haiku`, where link-chasing for an exact literal is least reliable.

The failure is silent. An unsigned post still succeeds and looks right to the poster; the comment is simply indistinguishable from a human one from then on, which breaks echo-protection and addressed-by-loop classification permanently.

## Fix

Ship the signature as `plugin/scripts/sign-body.sh` and pipe every posted body through it. Modes map onto the two markers already specified:

- `loop` — prefixes `<!-- muggle-do:bot -->`, the echo-detection token
- `editable` — prefixes the dedup marker and cuts the previous signature, so a description re-posted on every refresh keeps exactly one
- `plain` — no marker

Signing is idempotent, so refresh paths re-sign without checking.

All ten posting call sites now sign, including the four GitLab recipes that had no signing step and `reply-routing.md`, whose documented reply bodies carried no signature at all. The six duplicated copies of the literal are collapsed — the script is the only definition.

## Enforcement

`scripts/check-post-signatures.mjs`: a posting command inside a shell fence must be accompanied by a signing step. Prose naming a command (`never call gh pr comment here`) is a mention, not a call site, and reads through the same CLIs never match.

It joins backslash-continued lines, so a flag cannot hide on the next physical line — the gap that let `reply-routing.md`'s multi-line `gh api --method POST` slip past a naive per-line match.

Wired as the `post-signatures` CI job and `pnpm run verify:signatures`.

## Verification

- Linter against pre-fix `master`: **10 violations caught**, every one a real unsigned path. Clean on this branch.
- `sign-body.sh` exercised for all three modes, idempotency, and signature replacement.
- Full test suite green — 1281 tests across 87 files, including 14 new ones for the linter.
- `tsc --noEmit`, `check-skill-deps`, eslint all clean; `bash -n` on the script; LF endings confirmed.
- `build-plugin` artifact confirmed to ship `dist/plugin/scripts/sign-body.sh`, so `${CLAUDE_PLUGIN_ROOT}/scripts/` resolves at runtime.

One existing assertion in `pr-followup-tracking-flow.test.ts` moved with the endpoint it guards: `per-comment-replies.md` now delegates to the shared recipe, so the endpoint check reads it there and the nested-reply assertion stays put.

## Note

This is the fix `arm-watcher.md` already documents for the watch loop — *"writing the loop from this prose was how it drifted"* — applied to the signature, which never got it.

https://claude.ai/code/session_011HUYRMQkhHvRze25mKdMmv


plugin/skills/_shared/pr-followup-helpers/reply-routing.md
```diff
@@ -7,21 +7,25 @@ GitHub's PR APIs are not uniform across comment types. Route by parent type.
 A comment attached to a specific file:line that belongs to a review thread.
 
 ```bash
+body="$(printf 'Done in %s — renamed `fooBar` to `foo_bar`.\n' "$(git rev-parse --short HEAD)" \
```
stan4git: please make sure all your changes are VCS tool agnostic

stan4git: For vendor specific instruction , put them under the tool's instuction and reference here

stan4git: Addressed in a3b8d50: routed by parent type and moved every vendor command out of this file.

`reply-routing.md` now names the route (line-level reply, review-body-only, failing CI) and defers the command to the resolved provider's recipe — `github` to `vcs/github/reply-line-comment.md`, `gitlab` to `vcs/gitlab/reply-discussion.md` — with the review-body-only route marked GitHub-only, since GitLab has no review envelope and every note is already a discussion. `loop-signature.md` lost its GitHub-only framing the same way; the marker was always emitted on both providers.

PR #397, opened 2026-08-14
[UNIT-ONLY] Add a staging runtime target and staging release channel

stan4git:
## Goal

Give `@muggleai/works` a `staging` runtime target and a `staging` publish channel, so a harness build pointing at the Muggle AI staging backend can be produced and published without disturbing the default install.

Design: `muggle-ai-brain/architecture/2026-08-13-harness-staging-channel-design.md`. Plan: `muggle-ai-brain/pending/2026-08-13-harness-staging-channel-plan.md`.

## Acceptance criteria

- A runtime target selects prompt-service URL, Auth0 tenant, and studio release stream together — adding a target is one table entry, not a conditional at each site.
- `MUGGLE_MCP_PROMPT_SERVICE_TARGET` accepts `staging`; an unknown value is rejected naming the offender and listing the valid ones.
- Per-setting env overrides still win over the profile.
- `getConfig()` resolves on every target, including one whose Auth0 client is unprovisioned, so `status` and `doctor` keep working on a staging build.
- `muggle login` fails before contacting Auth0 when the active target has no device-code client, naming the target and what to provision.
- Studio checksums are recorded and verified per release stream.
- The publish workflow takes a `channel`; production is the default and behaves exactly as before.
- A staging publish bakes the staging target, versions a pre-release, and publishes under the `staging` dist-tag, leaving `latest` untouched.
- The pipeline refuses to publish if the baked target disagrees with the requested channel.

## Changes

**Runtime target model.** New `runtime-target-types.ts`, `runtime-target-constants.ts`, and `runtime-target.ts`. `config.ts` loses five `DEFAULT_*` constant pairs and four near-identical per-setting getters, and reads the active profile instead.

**Profile data in `config/runtime-targets.json`.** `postinstall.mjs` and `verify-electron-release-checksums.mjs` are plain JavaScript that cannot import the bundled TypeScript, yet both need the studio release tag prefix. One JSON keeps the runtime, the installer, and the release gate from drifting. Ships in the tarball.

**Checksums keyed by release stream, not by runtime target.** Dev and production install the same studio binary. Keying by target left `dev` with no recorded hashes, which would have silently skipped checksum verification on every dev install.

**Publish channel.** `publish-works-to-npm.yml` gains a `channel` input, a staging pre-release version step, a smoke-install assertion that the baked target matches the channel, and a dist-tag on publish.

### Two bugs found while wiring this up

- `postinstall.mjs` built the studio download URL from a hardcoded `electron-app-v` prefix, so a staging build would have downloaded the production studio.
- `config.ts` resolved its package root two levels up from `packages/mcps/src/shared`, landing on a `package.json` with no `muggleConfig` — so every config read threw outside a bundled build.

## Validation

`unit-only`. This repo ships a CLI, an MCP server, and skill files; there is no web surface for a browser to drive, so the E2E stage recorded a SKIPPED verdict with that reason rather than a `runId`.

`npm test` exit 0 — 90 files, 1309 passed, 27 skipped. `tsc --noEmit` exit 0. `lint:check` 0 errors.

Behavior exercised against the built artifact, not just the source:

| Command | Result |
| --- | --- |
| `MUGGLE_MCP_PROMPT_SERVICE_TARGET=production node bin/muggle.js status` | `Runtime target: production` / `https://promptservice.muggle-ai.com` |
| `...=staging ... status` | `Runtime target: staging` / `https://staging.promptservice.muggle-ai.com` |
| `...=staging ... login` | refuses with the provisioning message, before any Auth0 call |
| `...=prodction ... status` | `Invalid ... Expected one of: production, staging, dev.` |
| `node scripts/verify-electron-release-checksums.mjs` | verifies `electron-app-v1.8.0`; skips staging, none recorded |
| `npm pack --dry-run` | `config/runtime-targets.json` present in the tarball |

## Not in this PR

The staging studio release stream (teaching-service) and the deployment-ritual change (brain) are separate PRs. Until the studio stream exists, `checksumsByStream.staging` is empty and the verification script skips it.

**The staging channel cannot authenticate anyone yet.** The staging Auth0 tenant has no device-code client — confirmed by probing its device authorization endpoint: the client the staging backend validates against refuses the `device_code` grant, and the production client does not exist in that tenant. One must be provisioned there, then its ID recorded in `config/runtime-targets.json`. Everything here builds and publishes without it.

🤖 Opened by Muggle Works via `/muggle-do`


.github/workflows/publish-works-to-npm.yml
```diff
@@ -42,6 +42,14 @@ on:
           - patch
           - minor
           - major
+      channel:
+        description: "Release channel. production publishes to the latest dist-tag; staging publishes a pre-release under the staging dist-tag."
+        required: false
+        type: choice
+        default: production
```
stan4git: please default staging

stan4git: Done in `28d9248` — `channel` now defaults to `staging`, and the options are ordered staging-first so the safe one is also the one in front of you.

The reasoning matches yours: accepting the defaults on a hand-run was moving the `latest` dist-tag, so the riskiest path was the one requiring no thought.

One carve-out worth flagging, since it's a behavior difference rather than a default: a `v*` tag push carries no channel input and still resolves to production. Pushing a version tag is already an explicit release act, and the staging channel rewrites the version as a `-staging.N` pre-release — which a tag naming an exact version can't mean. Recorded in `contracts.json` as `defaultChannel: staging` / `tagPushChannel: production`, and commented at the input.

If you'd rather tag pushes also go to staging, say so — it means dropping the pre-release rewrite for that path, so it's a real change rather than a flag flip.

.github/workflows/publish-works-to-npm.yml
```diff
@@ -42,6 +42,19 @@ on:
           - patch
           - minor
           - major
+      channel:
+        # Defaults to staging so a hand-run of this workflow cannot move the
+        # latest dist-tag by accident. Publishing to production is a deliberate
+        # selection. A `v*` tag push carries no input and still resolves to
+        # production — pushing a version tag is already an explicit release act,
+        # and a tag cannot mean a pre-release version that contradicts it.
+        description: "Release channel. staging publishes a pre-release under the staging dist-tag; production publishes to the latest dist-tag."
```
stan4git: Can you actually make two publish pipeline, one prefix with "staging" and the other prefix with "production"

stan4git: Done in `da89efc` — same shape as the studio split.

- **`publish-works-to-npm-reusable.yml`** — verify, package-audit, smoke-install. Shared, channel-agnostic.
- **`publish-works-to-npm.yml`** — production. Filename and `name:` deliberately unchanged.
- **`staging-publish-works-to-npm.yml`** — staging: bakes the staging target, pre-release version, `--tag staging`.

The `channel` dropdown is gone. Which pipeline you run *is* the channel, which also supersedes the "default staging" fix — there's no default left to get wrong.

**One thing constrained the design, and it's the reason the split isn't cleaner.** The publish job is *not* in the reusable workflow. npm trusted publishing binds to the **calling workflow's filename** — this file's own header has warned about that since before I touched it. A shared publish job would give both channels one identity and break the production config npmjs.com already trusts. So each pipeline keeps its own ~20-line publish job; the ~200 shared lines are what moved.

**Needs you before the first staging publish:** npmjs.com to `@muggleai/works` to Access to Trusted Publishing must list `staging-publish-works-to-npm.yml` alongside the production one. Without it the staging run fails at the last step with ENEEDAUTH — after verify, package and smoke-install have all passed. Production is unaffected either way.

Two follow-ons I fixed rather than left dangling:
- The staging pipeline asserts after publishing that `latest` didn't move. That failure is silent otherwise, and the dist-tag is the only place it shows.
- `/deploy-all-prod` passed `-f channel=production`, an input that no longer exists — dispatching with an unknown input fails outright. Corrected in muggle-ai-brain#157, which was still open.

`npm test` exit 0, contracts verification passes, all three workflows parse, and I checked both callers wire to the shared build and keep the `npm-publish` environment.

.github/workflows/publish-works-to-npm.yml
stan4git, line 1: explictly name this pipeline and file "production"

.github/workflows/publish-works-to-npm.yml
stan4git: image
Is this what you mean?


There is already a OIDC for pipeline "publish-works-to-npm.yml", how to proceed?

stan4git: Yes, that's the right screen — and it answers the question in a way that kills the split. **Don't change anything on npmjs.com. Leave the existing entry alone.**

I checked the docs rather than guessing, because getting this wrong breaks production publishing:

> "Each package can only have one trusted publisher configured at a time."
>
> "validation checks the calling workflow's name instead of the workflow that actually contains the publish command"
>
> — https://docs.npmjs.com/trusted-publishers

Two consequences:

1. **One package, one trusted publisher.** Two publishing pipelines can't both hold the OIDC identity. Whichever filename isn't configured fails with ENEEDAUTH.
2. **The reusable-workflow escape doesn't exist.** npm validates the *caller*, not the workflow containing `npm publish` — so I can't give both channels one shared identity that way either.

So I've reverted the split in `5786e1e`. Publishing is one workflow again, selecting its channel by input, and **the filename is unchanged** — your existing OIDC entry keeps working with no reconfiguration.

That also drops the earlier "update npmjs.com before the first staging publish" item. It no longer applies.

Your naming request I can't satisfy for this workflow specifically: it publishes both channels, so calling it "production" would be wrong, and renaming it at all would invalidate the OIDC entry in your screenshot. The safety you were after is still there via the input defaulting to `staging` — the accident-prone path is still the one that requires no thought.

**The studio split in muggle-ai-teaching-service#356 stands** — that publishes GitHub releases through a GitHub App token, not npm OIDC, so the single-publisher limit doesn't apply. Two pipelines there, one here.

If you do want two npm pipelines, the only route is a granular access token for staging instead of OIDC. I'd advise against it: it trades provenance and a short-lived credential for a long-lived secret, to gain naming symmetry.

stan4git: Okay, no need for 2 pipelines. Would the "staging" deploy option offered in my screenshot works with your change?

stan4git: Yes — and better than that, **you don't need to touch npmjs.com at all.** Not for production, not for staging.

The existing entry already covers both channels, because after the revert both run through the same workflow and the same environment. I checked the three fields npm matches on, and all are byte-identical to what you have configured:

| Field | Value | Changed? |
| --- | --- | --- |
| Workflow filename | `publish-works-to-npm.yml` | no |
| Workflow `name:` | `Publish Works to npm` | no |
| Environment | `npm-publish` | no |

A staging publish is the *same* workflow run with `channel=staging`. The only thing that differs is the last line — `npm publish … --tag staging` instead of publishing to `latest`. OIDC has already been established by that point, identically, because it is the same job in the same file in the same environment.

That is the upside of the single pipeline I did not appreciate when I split it: one trusted publisher covers everything, and there is no second registry config to drift or be forgotten.

On the "staging" option in your screenshot — if that is the optional **Environment** field, leave it as `npm-publish`. Pointing it at a GitHub environment named `staging` would break *both* channels, because the publish job runs under `npm-publish` regardless of which channel it is publishing. The channel is a workflow input, not a GitHub environment.

So the npmjs.com item comes off your list entirely. What remains is the staging Auth0 device-code client, and reviews.

PR #404, opened 2026-08-18
feat(browser-bench): drive studio through the WebVoyager harness from a CLI

stan4git:
Adds the `runTaskAsync` seam `runBatchAsync` was written against: a tsx CLI that spawns one studio process per WebVoyager task, keeps each attempt on a fresh browser profile in its own trajectory directory, and renders the batch report.

## CLI

```
tsx internal/browser-bench/src/run.ts \
  [--tasks <path>] [--limit <n>] [--concurrency <n>] [--out <dir>] [--resume]
```

Results append to `<out>/partial.jsonl` as they land, so `--resume` skips what a previous run already recorded and folds those results back into the report in task-file order — a resumed batch reports identically to one that ran start to finish. Without `--resume` an existing log is deleted rather than merged into.

## Spawn contract

```
<studio-binary> --benchmark-task <path/to/task.json> --out <path/to/result.json>
```

Exit 0 means the attempt completed: `studioStatus: "success"` scores a pass, any other status a fail. A non-zero exit, a missing or unreadable result file, or outliving `TASK_TIMEOUT_MS` kills studio and throws, so the orchestrator records an infrastructure error and the pass-rate denominator stays honest. `tokensUsed` is `0` because studio does not report token spend yet; `judgeVerdict` stays unset for the judging pass.

The binary comes from `MUGGLE_STUDIO_BIN`, defaulting to `muggle-studio` on `PATH` — no machine-specific path in the tree. The per-task browser profile travels in `MUGGLE_STUDIO_BROWSER_PROFILE_DIR`, because the contract's two flags carry only the task and result paths and a profile has to reach studio somehow. Studio must honour it, or logged-in state leaks between tasks.

## Task data

`data/webvoyager-smoke.jsonl` carries 20 records copied verbatim from upstream `MinorJerry/WebVoyager` `data/WebVoyager_data.jsonl` (MIT): the first task of each of the 15 sites, plus a second from the first five. The full 643 are upstream's to publish and version — vendoring them would fork a dataset this repo does not own, and a smoke slice is what a code change needs to prove the harness still drives studio end to end. The README documents fetching the full set.

## Testing

Spawn and disk access are injected ports, so no test launches a binary or touches the filesystem. Covered: flag parsing and its rejections, the partial-log round trip including a line truncated by a killed run, the resume filter and result ordering, `result.json` to `TaskResult` mapping, and the runner's non-zero-exit, missing-result-file, spawn-failure, and timeout paths. The two-flag contract is pinned by its own assertion.

`npx vitest run internal/browser-bench` — 51 tests green across 9 files.
`npx tsc --noEmit` and `npx tsc -p internal/browser-bench --noEmit` — clean.
`npx eslint internal/browser-bench` — clean.

A real end-to-end run with `MUGGLE_STUDIO_BIN` pointed at a binary that does not exist produced the expected tree — `task.json` per task, empty profile directories, three infrastructure errors in `partial.jsonl`, and a rendered `report.md` — and a second run with `--resume` had nothing left to do.

https://claude.ai/code/session_018AzJ1mPrJ5YqZ4SEcqnZbT


stan4git:
Reconciling the six contract findings from the studio side. Two needed real changes, both now made in teaching-service#357 (`a840505`).

**3. `studioStatus` — this was a genuine bug, and the most valuable thing to come out of building the two halves separately.**

You inferred `"success"` → Pass from the `orchestrator.test.ts` fixture. That was right, but the studio was emitting `"active"` — the agent loop's internal spelling for success (`ActionScriptStatusEnum.Active`). So **every successful task would have scored as a failure**, and the first run would have reported near-0%: a catastrophic-looking capability result that was really a wiring bug.

Fixed on the studio side, not yours. It now stamps an explicit `BenchmarkStudioStatus` (`success` / `goal_not_achievable` / `failed`) rather than leaking an internal enum. Unrecognised values resolve to `failed` — a status the mapping has never seen is not evidence of success. Five tests pin it, including that no spelling other than `success` can ever come out as a pass.

Your mapping needs no change. Keep `"success"` → Pass.

**1. Browser profile — please drop the env var.**

The out-of-band `MUGGLE_STUDIO_BROWSER_PROFILE_DIR` isn't needed, and neither is a sixth task field. The studio already had this: `freshSession` clears cookies and local storage when the webview attaches. The benchmark context was just failing to set it, so it never fired.

It now sets `freshSession` itself rather than accepting a caller flag, because a benchmark task must always start clean and that shouldn't be something a caller can forget. Isolation is a property of benchmark mode.

So: **the contract stays at two flags and five task fields.** Please remove the env var so we don't carry a dead mechanism.

**Confirmed as-is, no change needed:**

- **4. `trajectoryDir` echoed both ways** — studio echoes what it was given; it never relocates. Taking studio's value is harmless.
- **5. Exit 0 with no `result.json`** — your conservative reading is correct. The studio only exits 0 after writing, so treating a missing file as an infrastructure error is right.
- **6. Windows `taskkill /T /F`** — correct and necessary. Electron spawns renderer and GPU children; `child.kill()` alone would strand them across a long batch.

**Open, not blocking the first run:**

- **2. No token field.** Agreed — `TaskResult.tokensUsed` exists so a score is reportable next to its cost, and the studio doesn't emit it yet. Worth adding before any *published* number, not before the first smoke run.
- **npm script** — deliberate gap, agreed; worth a separate small PR alongside `studio-gen-eval`'s.

PR #409, opened 2026-08-25
feat(guardrails): durable comment-reply ledger with claims and CAS writes

stan4git:
Implements `architecture/2026-08-25-comment-reply-ledger-design.md` (merged on brain `main`, amended by brain#165).

## Problem

A review round can apply the changes reviewers asked for, push, and never reply in-thread. The diff looks complete, the reviewer gets no answer where they asked, and the watcher re-dispatches the same thread forever because the `<!-- muggle-do:bot -->` marker that retires it only ever arrives on a reply.

The first attempt at this PR enforced it from whatever a session happened to observe. Review of that attempt found four holes, all now closed:

| Hole | Now |
| :-- | :-- |
| Marked on **attempt**, so a rejected reply still closed the obligation | Marked only when the provider returns a created comment id |
| **Session-scoped**, so a later session reprocessed answered comments | Durable per-PR ledger in the muggle-do slot |
| No **processing** state — a crashed round looked like one never started | Claim, with state derived from it |
| **Lost updates** — no lock, no CAS, four observers racing per tool call | Compare-and-swap under an exclusive lock, atomic replace |

That last one was **measured, not assumed**: instrumenting the four installed Bash observers showed all of them entering before any exits, in every trial, with completion order differing from start order. It is a live defect in the five gates that already ship, which `e5c7142` fixes for all of them.

## Design

**Thread-keyed, closed against a declared covered set.** Not a high-water mark — a mark cannot tell three comments answered by one reply from three answered by none. Every comment post-dating a thread's newest loop reply is its own obligation, so one reply may declare all three, while a reply addressing only the newest leaves the other two open and named. This also survives the handling-window race: a comment arriving mid-round is never in the covered set, where a mark re-derived after replies would swallow it unread.

**Claims expire on holder death, never a timer.** A round legitimately runs twenty minutes, so any expiry short enough to recover from a crash would expire out from under a slow worker and hand the thread to a second one — the duplicate processing this exists to prevent. A claim from another machine cannot be probed, so it falls back to a backstop measured in hours.

**The claim replaces the push signal.** A thread only enters the ledger when a round claims it, and claiming is what taking the work means — which also catches a round that answers a question with no code change, a case the push signal silently missed.

**Failure policy splits by store**, decided by the concurrency measurement: the session store fails open on contention (it sits in the critical path of every tool call), the ledger waits and retries (a dropped write is a missed obligation).

## Verification

- full suite green — 102 files, 1464 tests
- `tsc --noEmit` clean; eslint clean on touched sources; `check-skill-deps` OK across 366 cross-skill links
- end-to-end through the shipped bash wrappers: claim a thread → gate blocks naming the comment → confirmed reply → gate silent
- covered explicitly: three consecutive comments are three obligations; a provider-rejected reply leaves the obligation open; another session's claim never blocks this one
- E2E not applicable — Stop-hook plus skill-doc wiring, no browser surface

https://claude.ai/code/session_01H5x29dpkFsVvENqJDFeqxg


plugin/scripts/guardrail-comment-reply-gate.sh
stan4git, line 1: Would this keep writing to disk??

stan4git: Addressed in c420798: no — the wrapper only greps the state file, and the Node handler writes solely to bump the block count, so the gate's lifetime budget is 3 writes per session.

Full accounting, since "runs on every turn end" is the right thing to be suspicious of:

- **Wrapper** — pure reads. It greps `~/.muggle-ai/guardrails/<sid>.json` and returns `{}` in-shell. On a session with no review round the file is absent or those fields are unset, so Node never starts.
- **Gate** — `writeState` is reachable only on a `Block` decision. `MAX_REPLY_BLOCKS` (3) caps it; after that `Release` returns `{}` without writing.
- **Observer** (`record-comment-replies`, every Bash call) — every `apply*` returns the same object reference when nothing changed, and the caller writes only on `next !== state`. A repeated payload is a no-op.

c420798 pins both halves at the CLI entry rather than leaving it as my say-so: the gate blocks exactly 3 times and the state file is byte-identical across the released turns after, and a repeated observer payload leaves it untouched.

One thing your question surfaced that I did **not** change: after `Release` the shell pre-filter stops short-circuiting (`reviewWorkPushed` true, `commentRepliesOwed` non-empty, no skip recorded), so Node cold-starts once per turn end for the rest of the session, returning `{}` each time. No disk write, but not free either. All five gates share that shape (e2e, terminal, watch, walkthrough, debug), so the one-line fix — stamp `commentReplySkipped` on release so the pre-filter goes quiet — would make this one diverge from its siblings. Happy to do it here, or across all five as its own PR; say which.


stan4git: Follow-up on the above, in c88acd5: this PR is **held** rather than merged.

Your question about disk writes led to an audit that found two more holes I hadn't flagged: replies were marked on *attempt* rather than confirmed success (so a 4xx reply still marked the comment addressed — exactly the case `per-comment-replies.md` tolerates and logs), and the whole record was session-scoped, so a comment addressed in one session was invisible to the next.

The decision was to design the ledger properly before merging this. Spec is in multiplex-ai/muggle-ai-brain#163:

- durable per-PR ledger in the muggle-do slot, nag counters stay session-local
- thread-keyed, closed against a **declared covered set** rather than a high-water mark — a watermark can't tell three comments answered by one reply from three answered by none
- claims expire on holder **death**, not a timer — a wall-clock lease expires out from under a cycle that's merely slow, which is the duplicate-processing failure it's meant to prevent
- compare-and-swap under an `O_EXCL` lock, plus atomic replace in `sessionState.ts`, which closes the lost-update window for all five existing gates

The branch is rebased onto current master (a5a27e8) and green, so it stays mergeable while the design settles.

PR #431, opened 2026-08-27
fix(mcp): scripts summary is per test case, and is slimmed

stan4git:
**Stacked on #430** (the naming change) — review that first; this PR's diff is the last commit only.

Follow-up to muggle-ai-prompt-service#739, which stopped gating the scripts-summary endpoint on a test case having a script.

## The description was lying

It read *"Get a summary of test scripts for a project."* Since #739 the rows are one per **test case**: a hand-written case with no script yet gets a row too. An agent reading the old description counts rows as scripts and overcounts, or sees a row and concludes a script exists. The description now says rows are per test case, that `testScriptId` may be absent, and gives the response shape.

## The payload had no ceiling

The tool sent no pagination, so it took the legacy path and returned the **whole project** as a bare array of full `ITestSummary` rows — use case, test case state view, script, workflow run, studio result — and that array just grew from "rows = scripts" to "rows = all test cases".

It now sends `page`/`pageSize`/`sortBy`/`sortOrder`, and a `mapFromUpstream` transform keeps each row to the fields an agent reasons about, mirroring what `mapTestRunsSummary` already does for the runs tool.

**Pagination was not the optional half of this.** The transform reads an envelope, and without those params the endpoint answers with a bare array — the transform would have returned an empty page. Sending them is what makes slimming possible at all, and it moves the backend onto its cheaper Firestore-native path as a side effect. All four params carry zod defaults, so a caller passing only `projectId` still validates.

## Why a sibling transform rather than reusing the runs one

The row shape genuinely differs — script id instead of workflow run id, `scripts` instead of `runs` — so a shared generic would have fit neither. The new module keeps the runs transform's two documented decisions (no parallel TS input interface; no per-page aggregate) and adds one of its own:

> a script-less row is **kept**, with `testScriptId` omitted — never dropped, and never given an invented id. Surfacing those cases is the whole point of the endpoint.

That rule has its own test.

## Verification

23 suites / 250 tests pass (was 22 / 242). The new transform suite mirrors the runs one, including its size-reduction assertion — a heavy 20-row page slims to under 10% of its raw size. `tsc --noEmit` and eslint clean. The tool-surface snapshot is updated in this PR, as its own comment asks for.

The endpoint behaviour this depends on is verified live: against a backend carrying #739, the project's summary returns 19 rows against a `totalCount` of 19, including a hand-written DRAFT case with no script.

### DIFF
--- a/packages/mcps/src/mcp/e2e/contracts/index.ts
+++ b/packages/mcps/src/mcp/e2e/contracts/index.ts
@@ -483,6 +483,22 @@ export const ProjectTestResultsSummaryInputSchema = z.object({
 
 export const ProjectTestScriptsSummaryInputSchema = z.object({
   projectId: IdSchema.describe("Project ID (UUID) to get test scripts summary for"),
+  page: z.number().int().positive().default(1).describe("Page number, 1-based. Defaults to 1."),
+  pageSize: z
+    .number()
+    .int()
+    .positive()
+    .max(100)
+    .default(20)
+    .describe("Number of rows per page. Defaults to 20, max 100."),
+  sortBy: z
+    .enum(["createdAt", "updatedAt"])
+    .default("createdAt")
+    .describe("Sort field for the page. Defaults to createdAt (stable under concurrent writes)."),
+  sortOrder: z
+    .enum(["asc", "desc"])
+    .default("desc")
+    .describe("Sort direction. Defaults to desc (newest first)."),
 });
 
 export const ProjectTestRunsSummaryInputSchema = z.object({
--- a/packages/mcps/src/mcp/tools/e2e/test-scripts-summary-transform.ts
+++ b/packages/mcps/src/mcp/tools/e2e/test-scripts-summary-transform.ts
@@ -1,0 +1,96 @@
+/**
+ * Slim the per-row payload returned by the test-scripts/summary endpoint.
+ *
+ * The rows are one per TEST CASE, not one per script: a case that has never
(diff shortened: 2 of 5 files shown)


stan4git:
Note on CI: this PR shows **zero** checks, which is not the same as green. `ci.yml` triggers on `pull_request: branches: [master]`, and this PR's base is #430's branch, so no workflow has run against `f20578d` (0 check-runs by SHA).

That resolves itself — when #430 merges, GitHub retargets this to `master`, which fires the `pull_request` event and CI runs.

Verified locally in the meantime: `packages/mcps` at 23 suites / 250 tests, `tsc --noEmit` clean, eslint clean on every changed file.

stan4git:
⚠️ **This merged but did not reach `master`.** Re-landed as #432.

This PR's base was #430's branch. #430 merged first, which absorbed that branch into master — so merging this one afterwards put the commit on the now-dead base branch, not on master. GitHub shows MERGED either way.

Confirmed against `master`:
- `test-scripts-summary-transform.ts` → 404, absent
- `mapTestScriptsSummary` → 0 occurrences
- the stale `"Get a summary of test scripts for a project."` description → still present
- `git merge-base --is-ancestor f20578d origin/master` → no
- (#430's renames *did* land: 78 × `const toolInput = input as`)

#432 re-lands the identical content based on master, and will actually get CI — this PR never had a single check-run, because `ci.yml` only triggers on `pull_request: branches: [master]`.

PR #453, opened 2026-09-14
feat(onboarding): offer a first-run walkthrough that initializes preferences

stan4git:
Every first-time installer now gets offered a one-time walkthrough that explains what Muggle Test is and writes their preferences.

## The bug this fixes

The SessionStart hook runs `muggle setup` silently (`>/dev/null 2>&1`), and `setup.ts` seeds `DEFAULT_PREFERENCES` whenever `isFirstRun()` is true. That predicate is only "does `preferences.json` exist" — so the file appeared on session one, and from then on nothing could tell *the user chose these* from *nobody was ever asked*. The hook's own `"not configured"` branch has been unreachable the whole time.

Seeding and consenting are now separate events. Completion is stamped as `onboardingCompletedAt`, a top-level sibling of `preferences`, in the same shape `disclosureShownAt` already used.

**Existing users are never interrupted, with no version tracking.** Postinstall runs *before* any `muggle setup`, so a preferences file existing at that instant means the person has already used Muggle — the installer stamps them. A genuinely new install has no file yet, so the walkthrough fires.

## What was added

`muggle init`, in three modes over one engine:

| Mode | Purpose |
| :--- | :--- |
| `muggle init` | Interactive terminal wizard |
| `muggle init --json` | Emits the walkthrough for an agent to render as native pickers |
| `muggle init --apply <file>` | Persists a front-end's answers |

Front-ends report *what the user picked*, never resolved preference values — the mapping lives in the service, so the terminal and agent surfaces cannot drift. `--apply` is a single write, because the MCP tools and `@muggleai/telemetry` both read-modify-write this file with no lock between them.

The questions reuse the grouping `/mprefs` already had, extended to cover all 23 keys (previously 19 were reachable). The agent surface is `plugin/skills/muggle-preferences/ops/onboard.md`.

## Two related defects fixed

- **Defaults drift.** The hook hardcoded 18 defaults at `ask` while `DEFAULT_PREFERENCES` held 23 at `always` — skills and MCP tools disagreed about what a user had chosen, which would have made any promise the walkthrough makes untrue. Both now resolve through `plugin/config/preference-defaults.json`, pinned to `DEFAULT_PREFERENCES` by a test.
- **Disclosure copy.** The first-run notice told users to run `muggle preferences set telemetryEnabled false` — a command this CLI has never registered. The upstream description of what is collected is unchanged; only the closing instruction is replaced, with the two opt-outs that do work (`MUGGLE_TELEMETRY_DISABLED=1`, or top-level `telemetryEnabled: false`).

## Judgment calls worth a reviewer's eye

- **Unselected toggles map to `ask`, not `never`.** `never` hard-disables a feature (`autoE2ETest: never` = never run E2E again), which isn't what choosing caution means. One exception, added in the follow-up: a key whose recommended value is already `never` (only `verboseOutput`) stays `never`, since it is ungated and has no picker to prompt at.
- **"Ask me before everything" falls back to a key's default where `ask` isn't allowed.** `watcherLifetime` takes a duration, so there is nothing to ask.
- **Nine groups, not eight.** `watcherLifetime` needs a single-select and can't live in a toggle group. The eight original headers are unchanged.
- **The backfill is its own module.** `package.json` only publishes `scripts/postinstall.mjs`; an un-whitelisted import would have crashed every real npm install, so `scripts/onboarding-backfill.mjs` is added to `files`.

## Verification

`pnpm test` (1769 tests, coverage gate included), `lint:check`, `typecheck`, `build`, `smoke:cli` (11 commands, up from 10), `verify:contracts`, `verify:plugin`, `verify:skill-deps` — all exit 0.

No browser E2E: this package ships no web surface. Substituted with real subprocess runs of the shipped `bin/muggle.js` and the real `ensure-electron-app.sh` under isolated `HOME`/`USERPROFILE` directories:

- **New install** — backfill declines to stamp, hook offers the walkthrough, `--json` emits 9 groups covering 23 unique keys, `--apply` writes 23 preferences plus the stamp, hook then goes quiet.
- **Existing user** — backfill stamps; their `autoLogin=never`, `autoE2ETest=ask`, `llmEnv` and `disclosureShownAt` all survive untouched.
- **Skip cap** — offers 1 and 2 re-offer, offer 3 retires permanently, `preferences` stays `{}` throughout.

`src/test/scripts/ensure-electron-app-onboarding.test.ts` pins the regression in CI: it executes the real hook and asserts that seeded-defaults-without-a-stamp still triggers the walkthrough.

## Follow-up commit (`514ed0e`)

Impact analysis surfaced four things worth fixing before merge:

- **The offer could never retire if a user just ignored it.** `onboardingOfferCount` only advanced on an explicit recorded skip, so the cap depended on model compliance. The hook now counts its own injections against the same shipped cap. A missing preferences file also stopped being a special case — one code path resolves both, so a brand-new install gets the resolved preference line too, not just the directive.
- **The npm tarball gate didn't cover the new install-time dependency.** postinstall statically imports `onboarding-backfill.mjs` and the hook reads two files under `plugin/config/`; if any fell out of `files`, every install would throw at module load while the gate designed to catch that stayed silent.
- **Six skill docs still claimed an absent preference resolves to `ask`** — unreachable now that the line carries all 23 keys resolved. Two more duplicated the allowed-values list without the `autoE2ETest` / `watcherLifetime` overrides, one of which is printed verbatim to users. The configure picker reached 19 of 23 keys; now all 23.
- **`verboseOutput` is ungated, so `ask` was unrepresentable for it.** An unselected toggle whose default is already `never` now resolves to `never`.

Also removes `getPostInstallGuidance` (no callers, and a contradictory second answer to "what does a new user do first").


stan4git:
**`agent-gate-eval` red is pre-existing on `master`, not from this PR.**

The failing scenario is `visual-walkthrough-builder / mode-a-posts-comment-to-pr` (1/5 against a 0.8 threshold). This PR touches no agent definition and no part of `muggle-pr-visual-walkthrough`:

```
git diff --name-only origin/master HEAD | grep -iE "visual-walkthrough|agents/"   # no matches
```

That job has failed on `master` on every scheduled run for the past week — 2026-09-08 through 2026-09-14 — and it also **passed on this branch at 08:20 before failing at 08:31** with identical skill content, so it is non-deterministic rather than content-driven.

Every other check is green, and the local suite covering this change passes in full (test / lint / typecheck / build / smoke:cli / verify:contracts / verify:plugin / verify:skill-deps all exit 0).

Flagging rather than "fixing" it: a change to an agent this PR never touched, made to turn a flaky pre-existing red green, would be a fix against working code.

PR #458, opened 2026-09-17
[UNIT-ONLY] Isolate the skill-gate eval and right-size its trigger

stan4git:
## Goal

The skill-gate eval judges a skill on its own definition, runs only when a PR could actually change its verdict, and documents itself truthfully.

## Acceptance Criteria

- The eval no longer loads the developer's filesystem settings or this repo's plugin hooks.
- A support file under a skill directory no longer forces the full gate suite; a preference gate contract and the harness itself still do.
- The property that makes that safe is pinned by a test, not by a comment.
- The README stops contradicting itself about CI and cites the real pass bar.

## Changes

**Isolation.** `internal/skill-gate-eval/src/harness.ts` called `query()` without `settingSources`, so it loaded every filesystem settings source — including this repo's `.claude/settings.json`, which enables the muggle plugin and lets its Stop guardrails fire *inside* the eval. `agent-gate-eval` was explicitly isolated for exactly this in `63b9e77`; this harness was missed, and instead carried a workaround denying the production muggle MCP tools. That workaround stays as a second line of defence — the cause is now closed rather than patched.

**Trigger.** Any non-`SKILL.md` file under a skill directory forced the whole gate suite, on the stated grounds that its scope was unknown. It isn't unknown: a gate prompt is built from the skill's `SKILL.md` plus the gate contract, and the harness denies the `Read` tool, so a skill under eval cannot open anything else. A support file is invisible to every gate, and running ~40 minutes of probabilistic evals for one is cost with no signal — #456 paid it for a documentation edit.

That inference is only safe while both properties hold, so `internal/skill-gate-eval/test/system-prompt.test.ts` pins them: the prompt carries the skill body and the gate contract, and never a decoy support file placed in the same directory. If it ever fails, the scope rule has to widen with it — the test is what forces that conversation.

The decision itself moves out of inline YAML into `scripts/skill-eval-scope.sh`, which reads changed paths on stdin and prints `full` / `scoped <skills>` / `skip`. That makes the rule unit-testable without building a throwaway git repository per case; the workflow keeps the git plumbing.

| Change | Before | After |
| :-- | :-- | :-- |
| Support file under a skill dir | full suite | skip |
| Preference gate contract | full suite | full suite |
| Eval harness | full suite | full suite |
| `SKILL.md` body line | full suite | full suite |
| `SKILL.md` `model:` only | scoped | scoped |

**Docs.** The README claimed the eval was "never wired to CI" three lines above its own "CI (blocking)" section, and cited a ≥99% bar against an actual `PASS_THRESHOLD` of `0.8` — a bar nobody could meet, documented for a job that blocks every PR.

## Validation

`unit-only` — the eval harness ships no web UI surface.

- Suite green: 1701 tests, exit 0; `tsc --noEmit` clean; eslint 0 errors.
- The scope rule was exercised by hand through Git Bash across all seven cases in the table above before being wired into the workflow; its vitest coverage is bash-gated, so CI runs it on Ubuntu.
- `scripts/skill-eval-scope.sh` verified LF-only — a CRLF shell script exits 127 on a Linux runner.

## Origin

All three defects surfaced while shipping #456, not from a fresh sweep. The flaky red on that PR was independent of them and resolved as measurement noise.

https://claude.ai/code/session_01XCN29HzzsvMgpsVx3PhvTA


stan4git:
### Muggle AI — PR visual walkthrough

**E2E skipped** — muggle-ai-works ships an eval harness, a shell script and docs here — no web UI surface to drive. Verified by the unit suite (1701 tests, exit 0) plus a hand-run of the scope rule across all seven cases through Git Bash.

